PR #15, opened 2023-10-31
Add mai/matic deprecation proposal.

reem:
The avalanche portion still fails testing with the `CALLER_NOT_RISK_OR_POOL_ADMIN` error, presumably because the executor should be the guardian on avalanche? Let us know what the right fix here would be in the new system.


src/20231031_Multi_MAIMIMATICDeprecation20231031/AaveV3Optimism_MAIMIMATICDeprecation20231031_20231031.sol
```diff
@@ -0,0 +1,85 @@
...
+    override
+    returns (IAaveV3ConfigEngine.CollateralUpdate[] memory)
+  {
+    IAaveV3ConfigEngine.CollateralUpdate[]
+      memory collateralUpdate = new IAaveV3ConfigEngine.CollateralUpdate[](1);
+
+    collateralUpdate[0] = IAaveV3ConfigEngine.CollateralUpdate({
+      asset: AaveV3OptimismAssets.MAI_UNDERLYING,
+      ltv: EngineFlags.KEEP_CURRENT,
+      liqThreshold: 65_00,
```
rustboyar, line 55: forum post differs from the snapshot and says LT will be 0.3
probably makes sense to mention the change on the forum

rustboyar:
please merge main, do forge install and re-run tests for avalanche
+ add snapshot links

reem:
Rebased on main, updated config and added the snapshot url, and regenerated. This fixed the issue with avalanche but now the optimism payload fails during the default test when trying to run the e2eTest at the end of the proposal. Appears to be error `59` or `PRICE_ORACLE_SENTINEL_CHECK_FAILED` - not sure exactly what's going on here, could use your help understanding what to do next.

Zer0dot:
> Rebased on main, updated config and added the snapshot url, and regenerated. This fixed the issue with avalanche but now the optimism payload fails during the default test when trying to run the e2eTest at the end of the proposal. Appears to be error `59` or `PRICE_ORACLE_SENTINEL_CHECK_FAILED` - not sure exactly what's going on here, could use your help understanding what to do next.

Hey try updating the fork block to 111887045, it seems the network may have been down for the selected block.

src/20231031_Multi_MAIMIMATICDeprecation20231031/AaveV3Optimism_MAIMIMATICDeprecation20231031_20231031.sol
Zer0dot: @reem

reem: Thanks, we've updated the forum post to match

reem:
Updated blocks and re-ran tests, that worked thanks!

Zer0dot (dismissed):

rustboyar (dismissed):

reem:
Added deployed addresses for all chains.

To confirm, with gov 2.5 and beyond avalanche payloads are now executed by normal governance and not by a steward?

reem:
Updated blocks and included requested parameter changes, as detailed in the updated proposal description.

reem:
I tried to merge main and update for changes to the generator, but I got some errors when trying to regenerate with this config, is there a summary of changes to the config that I should apply here?

reem:
The payloads will also need to be re-deployed after further review

reem:
Added newly redeployed payload addresses

sakulstra (approved):

reem:
To limit governance fatigue and maximize the voting period during the weekday, we will push this AIP on chain on Monday Nov 20.

PR #25, opened 2023-11-01
fix: better number formatting

sakulstra:
- closes https://github.com/bgd-labs/aave-proposals-v3/issues/4

### DIFF
--- a/generator/prompts.ts
+++ b/generator/prompts.ts
@@ -24,7 +24,6 @@ export function transformNumberToPercent(value: string) {
     return (
       new Intl.NumberFormat('en-us', {
         maximumFractionDigits: 2,
-        minimumFractionDigits: 2,
       }).format(value as unknown as number) + ' %'
     );
   }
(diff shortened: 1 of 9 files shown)


src/20231029_AaveV3Ethereum_ACIPhaseII/ACIPhaseII_20231029.s.sol
```diff
@@ -18,7 +18,7 @@ contract DeployEthereum is EthereumScript {
     IPayloadsControllerCore.ExecutionAction[]
       memory actions = new IPayloadsControllerCore.ExecutionAction[](1);
     actions[0] = GovV3Helpers.buildAction(address(payload0));
-
+    uint256 a = 10;
```
sakulstra, line 21:
```suggestion
```

sendra (dismissed):

generator/prompts.ts
```diff
@@ -19,29 +19,39 @@ function isAddressOrKeepCurrent(value: string) {
...
+    );
   }
   return value;
 }
 
-function transformNumberToHumanReadable(value: string) {
+export function transformNumberToHumanReadable(value: string) {
   if (value && isNumber(value)) {
-    return value.replace(/(?=(\d{3}$)+(?!\d))/g, '.');
+    return new Intl.NumberFormat('en-us').format(BigInt(value));
```
brotherlymite, line 35: When the value is a decimal (inputted via the keyboard) and when trying to convert it into `BigInt()` it throws an error and reverts. Maybe a good idea to validate that the number is not a decimal before and handle it gracefully.

sakulstra: @brotherlymite fair point.

Instead of doing this though i went ahead and created a new input prompt that allows specifying a pattern that must be followed (so numberInput will only allow numbers, percentInput only allows a single decimal point,... more to come i guess)

sakulstra: @brotherlymite fair point.

Instead of doing this though i went ahead and created a new input prompt that allows specifying a pattern that must be followed (so numberInput will only allow numbers, percentInput only allows a single decimal point,... more to come i guess)

brotherlymite: Thx, looks good now

brotherlymite (approved):

sendra (approved):

PR #31, opened 2023-11-02
refactor: refactor address input

sakulstra:


generator/features/eModesUpdates.ts
```diff
@@ -43,7 +47,9 @@ async function subCli(pool: PoolIdentifier) {
   if (shouldAddNewCategory) {
     let more: boolean = true;
     while (more) {
-      answers.push(await fetchEmodeCategoryUpdate(true));
+      const eModes = getEModes(pool as any);
+      const highestEmode = Math.max(...Object.values(eModes));
+      answers.push(await fetchEmodeCategoryUpdate(highestEmode + 1, true));
```
brotherlymite, line 52:
```suggestion
      const newEmode = Object.keys(eModes).length != 0 ? highestEmode + 1 : 0;
      answers.push(await fetchEmodeCategoryUpdate(newEmode, true));
```
I would also check if no eModes exists, with something like above - Otherwise `highestEmode` would return `-Infinity` when no emode exists.

sakulstra: solved it a bit differently, but good catch 👍

brotherlymite (approved):

sendra (approved):

PR #98, opened 2023-11-27
Add gauntlet CRV update proposal.

reem:
Had to specify the current reserve factors explicitly otherwise I received an invalid reserve factor error during tests, which is unusual.


brotherlymite:
> Had to specify the current reserve factors explicitly otherwise I received an invalid reserve factor error during tests, which is unusual.

Using KEEP_CURRENT on the reserve factor has some invalid condition on the Engine, we will fix that in the next update of the Engine. Using current reserve factor explicitly should be fine for now.

reem:
Added deployed payload addresses

brotherlymite (dismissed):

src/20231127_Multi_GauntletRecommendationToReactivateCRVBorrowingOnV3/GauntletRecommendationToReactivateCRVBorrowingOnV3_20231127.s.sol
```diff
@@ -0,0 +1,78 @@
...
+
+    // compose actions for validation
+    IPayloadsControllerCore.ExecutionAction[]
+      memory actionsEthereum = new IPayloadsControllerCore.ExecutionAction[](1);
+    actionsEthereum[0] = GovV3Helpers.buildAction(0x78acec1b92df516da14b3dbed91219772180330);
+    payloads[0] = GovV3Helpers.buildMainnetPayload(vm, actionsEthereum);
+
+    IPayloadsControllerCore.ExecutionAction[]
+      memory actionsPolygon = new IPayloadsControllerCore.ExecutionAction[](1);
+    actionsPolygon[0] = GovV3Helpers.buildAction(0x78acec1b92df516da14b3dbed91219772180330a);
```
brotherlymite, line 64: @reem can you fix the checksum of the address

brotherlymite (approved):

PR #100, opened 2023-11-29
Aave <> Gauntlet Renewal Proposal

reem:
Creating this ahead of time, this is only if the community approves the snapshot, but we wanted to share the payload in advance.


rustboyar:
> Thanks - in the forums, we commented [here](https://governance.aave.com/t/arfc-gauntlet-aave-renewal-2023/15380/9) that the payment will be in aUSDT instead of aUSDC. But when the AIP is published, we can remind the community again that it will be aUSDT instead of aUSDC.

ah, it's fine then

reem:
Added deployed payload address. Thanks for the help with verification.

brotherlymite (approved):

PR #126, opened 2023-12-11
Treasury Management - Polygon Migration v2 to v3

efecarranza:
# CHANGELOG

Withdraw from v2 polygon assets.
Deposit withdrawn assets into Polygon v3.

AIP.
Tests.

### DIFF
--- a/src/20231208_Multi_TreasuryManagementPolygonV2ToV3Migration/AaveV2Polygon_TreasuryManagementPolygonV2ToV3Migration_20231208.sol
+++ b/src/20231208_Multi_TreasuryManagementPolygonV2ToV3Migration/AaveV2Polygon_TreasuryManagementPolygonV2ToV3Migration_20231208.sol
@@ -106,24 +106,5 @@ contract AaveV2Polygon_TreasuryManagementPolygonV2ToV3Migration_20231208 is
       address(AaveV3Polygon.COLLECTOR),
       0
     );
-
-    AaveV2Polygon.COLLECTOR.transfer(
-      AaveV2PolygonAssets.LINK_UNDERLYING,
-      address(this),
-      IERC20(AaveV2PolygonAssets.LINK_UNDERLYING).balanceOf(address(AaveV2Polygon.COLLECTOR))
-    );
-
-    uint256 amountLink = IERC20(AaveV2PolygonAssets.LINK_UNDERLYING).balanceOf(address(this));
-    IERC20(AaveV2PolygonAssets.LINK_UNDERLYING).forceApprove(
-      address(AaveV3Polygon.POOL),
-      amountLink
-    );
-
-    AaveV3Polygon.POOL.deposit(
-      AaveV2PolygonAssets.LINK_UNDERLYING,
-      amountLink,
-      address(AaveV3Polygon.COLLECTOR),
-      0
-    );
   }
 }
--- a/src/20231208_Multi_TreasuryManagementPolygonV2ToV3Migration/AaveV2Polygon_TreasuryManagementPolygonV2ToV3Migration_20231208.sol
+++ b/src/20231208_Multi_TreasuryManagementPolygonV2ToV3Migration/AaveV2Polygon_TreasuryManagementPolygonV2ToV3Migration_20231208.sol
@@ -1,0 +1,129 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.0;
+
+import {IERC20} from 'solidity-utils/contracts/oz-common/interfaces/IERC20.sol';
+import {SafeERC20} from 'solidity-utils/contracts/oz-common/SafeERC20.sol';
+import {AaveV2Polygon, AaveV2PolygonAssets} from 'aave-address-book/AaveV2Polygon.sol';
+import {AaveV3Polygon, AaveV3PolygonAssets} from 'aave-address-book/AaveV3Polygon.sol';
+import {IProposalGenericExecutor} from 'aave-helpers/interfaces/IProposalGenericExecutor.sol';
+
+library AddressesToMigrate {
+  function getUnderlyingAddresses() internal view returns (address[8] memory) {
+    return [
+      AaveV2PolygonAssets.USDC_UNDERLYING,
+      AaveV2PolygonAssets.DAI_UNDERLYING,
+      AaveV2PolygonAssets.USDT_UNDERLYING,
+      AaveV2PolygonAssets.WETH_UNDERLYING,
+      AaveV2PolygonAssets.WMATIC_UNDERLYING,
+      AaveV2PolygonAssets.WBTC_UNDERLYING,
+      AaveV2PolygonAssets.GHST_UNDERLYING,
+      AaveV2PolygonAssets.LINK_UNDERLYING
+    ];
+  }
+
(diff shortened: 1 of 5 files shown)


src/20231208_Multi_TreasuryManagementPolygonV2ToV3Migration/TreasuryManagementPolygonV2ToV3Migration.md
```diff
@@ -0,0 +1,57 @@
...
+- [USDC](https://polygonscan.com/token/0x2791bca1f2de4661ed88a30c99a7a9449aa84174?a=0xe8599F3cc5D38a9aD6F3684cd5CEa72f10Dbc383)
+
+## References
+
+- Implementation: [AaveV2Polygon](https://github.com/bgd-labs/aave-proposals-v3/blob/main/src/20231208_Multi_TreasuryManagementPolygonV2ToV3Migration/AaveV2Polygon_TreasuryManagementPolygonV2ToV3Migration_20231208.sol), [AaveV3Polygon](https://github.com/bgd-labs/aave-proposals-v3/blob/main/src/20231208_Multi_TreasuryManagementPolygonV2ToV3Migration/AaveV3Polygon_TreasuryManagementPolygonV2ToV3Migration_20231208.sol)
+- Tests: [AaveV2Polygon](https://github.com/bgd-labs/aave-proposals-v3/blob/main/src/20231208_Multi_TreasuryManagementPolygonV2ToV3Migration/AaveV2Polygon_TreasuryManagementPolygonV2ToV3Migration_20231208.t.sol), [AaveV3Polygon](https://github.com/bgd-labs/aave-proposals-v3/blob/main/src/20231208_Multi_TreasuryManagementPolygonV2ToV3Migration/AaveV3Polygon_TreasuryManagementPolygonV2ToV3Migration_20231208.t.sol)
+- [Snapshot](https://snapshot.org/#/aave.eth/proposal/0x1b816c12b6f547a1982198ffd0e36412390b05828b560c9edee4e8a6903c4882)
```
sakulstra, line 48: i'm a bit confused with discussion & snapshot - both are quite old and talk about using fusion to swap assets etc.
This proposal seems to be doing sth completely different.

If this is not the wrong link I think at least clarifying in an update post what is done & why would be a good idea.

src/20231208_Multi_TreasuryManagementPolygonV2ToV3Migration/AaveV2Polygon_TreasuryManagementPolygonV2ToV3Migration_20231208.t.sol
```diff
@@ -0,0 +1,105 @@
...
+      AaveV3PolygonAssets.GHST_A_TOKEN,
+      AaveV3PolygonAssets.LINK_A_TOKEN
+    ];
+
+    uint256[] memory v3AssetsBalances = new uint256[](8);
+    uint256[] memory v2Assetsbalances = new uint256[](8);
+
+    for (uint256 i = 0; i < TO_MIGRATE.length; ++i) {
+      uint256 balanceV2 = IERC20(A_TOKENS[i]).balanceOf(address(AaveV2Polygon.COLLECTOR));
+      uint256 balanceV3 = IERC20(V3_A_TOKENS[i]).balanceOf(address(AaveV3Polygon.COLLECTOR));
```
eboadom, line 49: It works, but it is better to use `AaveV3Polygon.COLLECTOR` always, instead of mixing up v2 and v3, because they are the same address.
As in this case the objective is to check balances' consistency post-proposal, to be double sure you can just do an asset that both collectors are the same address.

This also applies to the payload, where you use sometimes v2 and others v3 for what it seems arbitrary reasons

src/20231208_Multi_TreasuryManagementPolygonV2ToV3Migration/AaveV2Polygon_TreasuryManagementPolygonV2ToV3Migration_20231208.t.sol
```diff
@@ -0,0 +1,105 @@
...
+      AaveV3PolygonAssets.WMATIC_A_TOKEN,
+      AaveV3PolygonAssets.WBTC_A_TOKEN,
+      AaveV3PolygonAssets.GHST_A_TOKEN,
+      AaveV3PolygonAssets.LINK_A_TOKEN
+    ];
+
+    uint256[] memory v3AssetsBalances = new uint256[](8);
+    uint256[] memory v2Assetsbalances = new uint256[](8);
+
+    for (uint256 i = 0; i < TO_MIGRATE.length; ++i) {
```
eboadom, line 47: The `++i` is fine and no need to change, but I recommend you to always to the "standard" `i++`.
Not sure if there are some weird gas savings, but going out of styling norms could create you big troubles on other projects.

src/20231208_Multi_TreasuryManagementPolygonV2ToV3Migration/TreasuryManagementPolygonV2ToV3Migration.md
```diff
@@ -0,0 +1,57 @@
+---
+title: "Treasury Management - Polygon v2 to v3 Migration"
+author: "TokenLogic"
+discussions: "https://governance.aave.com/t/arfc-migrate-consolidate-polygon-treasury/12248"
+---
+
+## Simple Summary
+
+This AIP will migrate all the DAOs assets held on Polygon v2 to Polygon v3.
```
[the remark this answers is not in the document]
eboadom: An additional comment on this. If your intention is to migrate all, most probably you should have a better test to actually go over all the v2 assets held by treasury and properly check that balance is minimal after migration

src/20231208_Multi_TreasuryManagementPolygonV2ToV3Migration/AaveV2Polygon_TreasuryManagementPolygonV2ToV3Migration_20231208.sol
```diff
@@ -0,0 +1,110 @@
...
+
+      AaveV2Polygon.POOL.withdraw(underlying, type(uint256).max, address(this));
+
+      uint256 amount = IERC20(underlying).balanceOf(address(this));
+      IERC20(underlying).forceApprove(address(AaveV3Polygon.POOL), amount);
+
+      AaveV3Polygon.POOL.deposit(underlying, amount, address(AaveV3Polygon.COLLECTOR), 0);
+    }
+
+    AaveV2Polygon.COLLECTOR.transfer(
```
eboadom, line 72: This should be done before the loop, and you are duplicated deposits on v3 for no reason

src/20231208_Multi_TreasuryManagementPolygonV2ToV3Migration/TreasuryManagementPolygonV2ToV3Migration.md
Matt-Graham-TL: This is the third attempt to migrate these assets from v2 to v3, see two relevant links below:

https://app.aave.com/governance/proposal/303/
https://app.aave.com/governance/proposal/335/

A comment has been added to the forum, linking this PR in the comments and highlighting the revised implementation methodology. 

https://governance.aave.com/t/arfc-migrate-consolidate-polygon-treasury/12248/4

PR #169, opened 2024-01-18
GHO Stability Module Launch

miguelmtzinf:


src/20240119_Gho_GHOStabilityModule/Gho_GHOStabilityModule_20240119.sol
```diff
@@ -0,0 +1,124 @@
...
+    IERC20(AaveV3EthereumAssets.LINK_UNDERLYING).forceApprove(
+      ROBOT_OPERATOR,
+      TOTAL_LINK_AMOUNT_KEEPERS
+    );
+
+    IAaveCLRobotOperator(ROBOT_OPERATOR).register(
+      'GSM USDC OracleSwapFreezer',
+      GSM_USDC_ORACLE_SWAP_FREEZER,
+      5000000, // gasLimit
+      LINK_AMOUNT_ORACLE_FREEZER_KEEPER
```
brotherlymite, line 115: Funding the keeper with 10 Link would not be sufficient for gasLimit of 5m as the keeper will then be underfunded and would not perform any action (this is unless you wish to fund it yourself later and do not wish for the DAO to pay)

brotherlymite: For reference, with the current gasPrice of 23 gwei and gasLimit of 5m the keeper requires a minimum link balance of 42 LINK

miguelmtzinf: Adjusted the gasLimit to 150k, which is the double of the gas cost for the keep action, and increased the LINK amount to 100 ether. Based on your example, we would have enough LINK for almost 80 keep actions on each

miguelmtzinf: decreased to 80 LINK each, so there is enough funds in the Treasury to pull

src/20240119_Gho_GHOStabilityModule/GHOStabilityModule.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Specification
+
+The proposed payload entails the comprehensive activation of GSM USDC and GSM USDT, involving the following steps:
+
+1. Incorporate GSM USDC and GSM USDT as facilitators of the GHO Token on Ethereum.
+2. Adjust the Fee Strategy for both GSMs to implement a 2% flat fee for both directions (buy/sell).
+3. Add both GSMs to the GSM Registry.
+4. Designate OracleSwapFreezer contracts as SwapFreezer entities in each GSM contract, respectively.
+5. Activate these OracleSwapFreezer contracts as keepers of the Aave DAO through AaveRobot with a funding of 10 LINK for each.
```
brotherlymite, line 36:
```suggestion
5. Activate these OracleSwapFreezer contracts as keepers of the Aave DAO through AaveRobot with a funding of 80 LINK for each.
```

miguelmtzinf: good point, thanks!

src/20240119_Gho_GHOStabilityModule/GHOStabilityModule.md
```diff
@@ -0,0 +1,73 @@
...
+- **Last Resort Liquidation**: Aave DAO is the exclusive entity granted with the role of last resort liquidation, empowering it to take control of GSM funds in worst-case scenarios.
+- **Swap Freeze**: Aave DAO and a chainlink-automated keeper contract have the authority to freeze the swap functionality. The chainlink-automated keeper contract bases its actions on the price of the exogenous asset, freezing if the price is outside the range and unfreezing if inside the range.
+- **Capital Allocation**: Supporting this feature by allowing ERC4626 assets as underlying assets. This enables redirecting the yield generated by the ERC4626 asset, while residing in the GSM contract, to the GHO Treasury.
+
+## Specification
+
+The proposed payload entails the comprehensive activation of GSM USDC and GSM USDT, involving the following steps:
+
+1. Incorporate GSM USDC and GSM USDT as facilitators of the GHO Token on Ethereum.
+2. Adjust the Fee Strategy for both GSMs to implement a 2% flat fee for both directions (buy/sell).
```
brotherlymite, line 33:
```suggestion
2. Adjust the Fee Strategy for both GSMs to implement a 0.2% flat fee for both directions (buy/sell).
```
Should be 0.2% no?

miguelmtzinf: yes, tysm :1st_place_medal:

eboadom:
Doesn't really belong to the proposal stage, but are there tests checking 1) the roles that are not touched on the proposal are held by expected addresses (e.g. Executor lvl1 of gov) and complementary 2) no other addresses keep those roles (from deployer for example).

src/20240119_Gho_GHOStabilityModule/Gho_GHOStabilityModule_20240119.sol
```diff
@@ -0,0 +1,126 @@
...
+  ) external returns (uint256);
+}
+
+/**
+ * @title GHO Stability Module
+ * @author Aave labs (@aave)
+ * @dev This proposal enables 2 GHO Stability Modules (USDC, USDT):
+ * - Addition of USDC and USDT GSMs as GHO Facilitators
+ * - Give Swap Freezer permissions to OracleSwapFreezers, one per module
+ * - Install a 2% fee strategy into both modules
```
eboadom, line 45:
```suggestion
 * - Install a 0.2% fee strategy into both modules
```

eboadom:
Is the `LIQUIDATOR_ROLE` set?

miguelmtzinf:
Just added tests to check permissions after proposal and fix the typo.

> Is the `LIQUIDATOR_ROLE` set?

`LIQUIDATOR_ROLE` and `TOKEN_RESCUER_ROLE` are not set because are not initially needed. Given that the DAO is the `DEFAULT_ADMIN` of the contracts, it can enroll itself as one of them. Same for `SWAP_FREEZER_ROLE`, but I enrolled the lvl1 executor as freezer so it's easy to use in case of need.

eboadom:
> Just added tests to check permissions after proposal and fix the typo.
> 
> > Is the `LIQUIDATOR_ROLE` set?
> 
> `LIQUIDATOR_ROLE` and `TOKEN_RESCUER_ROLE` are not set because are not initially needed. Given that the DAO is the `DEFAULT_ADMIN` of the contracts, it can enroll itself as one of them. Same for `SWAP_FREEZER_ROLE`, but I enrolled the lvl1 executor as freezer so it's easy to use in case of need.

eboadom (approved):

brotherlymite (approved):

PR #181, opened 2024-01-24
Fix text md file of GSM Launch

miguelmtzinf:


sakulstra (commented):
why the change? (seems to be just linting, right?)

anyhow you need to replace the git hash inside References back to main as otherwise it will not be updated again

sakulstra (approved):

PR #193, opened 2024-01-31
fix: generator bnb patches

sakulstra:


foundry.toml
```diff
@@ -26,17 +26,17 @@ bnb = "${RPC_BNB}"
 scroll="${RPC_SCROLL}"
 
 [etherscan]
-mainnet = { key="${ETHERSCAN_API_KEY_MAINNET}", chainId=1 }
```
sakulstra, line 29: seems like chainId is wrong: https://book.getfoundry.sh/reference/config/etherscan not sure if was changed or has always been wrong

generator/utils/importsResolver.ts
```diff
@@ -17,17 +17,17 @@ const GovernanceImports = [
 function generateAddressBookImports(code: string) {
   const imports: string[] = [];
   let root = '';
-  const addressBookMatch = code.match(/(AaveV[2..3][A-Z][a-z]+)\./);
+  const addressBookMatch = code.match(/(AaveV[2..3][A-Za-z]+)\./);
```
sakulstra, line 20: the regex expects at least one lowercase, but in case on BNB there is non. This fixes it.

sendra (approved):

PR #195, opened 2024-01-31
Polygon RF Updates Jan 31st

efecarranza:
# Changelog

Polygon RF Updates.


src/20240125_AaveV2Polygon_ReserveFactorUpdatesJan312024/ReserveFactorUpdatesJan312024.md
```diff
@@ -0,0 +1,40 @@
+---
+title: "Reserve Factor Updates (Jan 31, 2024))"
+author: "karpatkey_TokenLogic"
+discussions: "https://governance.aave.com/t/arfc-reserve-factor-updates-polygon-aave-v2/13937/14"
+---
+
+## Simple Summary
+
+This AIP is a continuation of Polygon ProposalID 27 and increases the Reserve Factor (RF) for assets on Polygon v2 by 5.00%, up to a maximum of 99.99%.
```
sakulstra, line 9: this is a bit weird reference imo... what you mean with 27 is the payloadId and it's a bit arbitrary as noone keeps track of them 😅 Better say it's a continuation of proposal 9 on governance v3 https://vote.onaave.com/proposal/?proposalId=9&ipfsHash=0x3dffd2f1b7a87a3f44ab31d0da89e5b64a6258468843c4599edd4527a4cd8c3a as i think proposal 9 was the last one of that type.

lib/aave-helpers
sakulstra, line 1: why is this updated?

src/20240125_AaveV2Polygon_ReserveFactorUpdatesJan312024/ReserveFactorUpdatesJan312024.md
efecarranza: ah that makes sense, thank you

lib/aave-helpers
efecarranza: removed! i guess i mistakenly updated when merging main

sakulstra (approved):

PR #231, opened 2024-03-05
feat: Add CCC update and new Bridge adapters for aDI

sendra:


sendra:
Closing as we will create separate proposals for each adapter

PR #244, opened 2024-03-12
stablecoin harmonization - review

Rozengarden:
my test are failling during the first call to vm.serializeJson(reserves, {})
with  error "Invalid Data"


eboadom:
Regarding assets (not so technical, but consistency-wise), the proposal is consistent with the table on Snapshot, but considering certain assets as "long-tail" collaterals while the following not seems completely inconsistent:

**Ethereum**
- Freezing and reducing LT of UNI, but not doing same for 1INCH, SNX, CRV, ENS. High-level feels like simply UNI should not be part of any off-boarding, being perfectly healthy. Very debatable BAL too, but in that case the size is too small, so legit.

**Polygon**
- GHST not included in freezing and LT reduction.

brotherlymite:
For the assets being freezed, I think it makes sense to change the LTV to 0 as well. 
In v3.1.0, it should happen by default when an asset is being freezed - but for now I think setting it manually makes sense.

For alternative stablecoins except USDC, USDT are we very strict to set LTV 0 and LT reduction by 3%?
If so, 
- No changes to sUSD on optimism?
- No reserve factor update on EURe on gnosis? also sDAI

src/20240312_Multi_StablecoinHarmonization/StablecoinHarmonization.md
```diff
@@ -0,0 +1,67 @@
...
+1. Focus on USDT and USDC as primary stablecoin collaterals.
+2. Set LTV to zero for alternative stablecoins and long tail assets.
+3. Enable a 3%/week LT reduction for these stablecoins assets, providing ample time for users to react and keep them as passive yield sources or borrowable assets that represent 99% of current user demand and most of protocol revenue.
+4. Propose the following adjusted standardized ReserveFactor categories for Aave assets:
+
+| Collateral Type                            | Reserve Factor % |
+| ------------------------------------------ | ---------------- |
+| wETH                                       | 15%              |
+| USDC/USDT                                  | 10%              |
+| Alternative stablecoins & long tail assets | 20%              |
```
brotherlymite, line 31: I would mention with the exception of MAI, where RF is `95%`

marczeller: updated

marczeller:
> For the assets being freezed, I think it makes sense to change the LTV to 0 as well. In v3.1.0, it should happen by default when an asset is being freezed - but for now I think setting it manually makes sense.
> 
> For alternative stablecoins except USDC, USDT are we very strict to set LTV 0 and LT reduction by 3%? If so,
> 
> * No changes to sUSD on optimism?
> * No reserve factor update on EURe on gnosis? also sDAI

sUSD implemented RF but no offboarding as there's a use case with synthetix there
EURe RF implemented but sDAI can't be borrowed so no implementation to do

eboadom:
@marczeller I assume keeping the LT reduction of UNI and BAL is on purpose, even if this https://governance.aave.com/t/arfc-stablecoin-harmonization-and-asset-parameters-optimization/16802/7 ?

marczeller:
> @marczeller I assume keeping the LT reduction of UNI and BAL is on purpose, even if this https://governance.aave.com/t/arfc-stablecoin-harmonization-and-asset-parameters-optimization/16802/7 ?

that is correct. we will add an option in snapshot vote to revert this, not change anything or confirm LT reduction allowing governance to decide.

eboadom (approved):

PR #305, opened 2024-04-18
Chaos Labs Engagement Amendment 20240415

eyalovadya:
Discussion: https://governance.aave.com/t/arfc-chaos-labs-engagement-amendment/17324

### DIFF
--- a/src/20240415_AaveV3Ethereum_ChaosLabsEngagementAmendment/AaveV3Ethereum_ChaosLabsEngagementAmendment_20240415.sol
+++ b/src/20240415_AaveV3Ethereum_ChaosLabsEngagementAmendment/AaveV3Ethereum_ChaosLabsEngagementAmendment_20240415.sol
@@ -11,27 +11,16 @@ import {IProposalGenericExecutor} from 'aave-helpers/interfaces/IProposalGeneric
  */
 contract AaveV3Ethereum_ChaosLabsEngagementAmendment_20240415 is IProposalGenericExecutor {
   address public constant CHAOS_LABS_TREASURY = 0xbC540e0729B732fb14afA240aA5A047aE9ba7dF0;
-  uint256 public constant STREAM_AMOUNT_GHO = 200_000 ether;
-  uint256 public constant STREAM_AMOUNT_AUSDC = 200_000e6;
+  uint256 public constant STREAM_AMOUNT_GHO = 400_000 ether;
   uint256 public constant STREAM_DURATION = 365 days;
   uint256 public constant ACTUAL_STREAM_AMOUNT_GHO =
     (STREAM_AMOUNT_GHO / STREAM_DURATION) * STREAM_DURATION;
-  uint256 public constant ACTUAL_STREAM_AMOUNT_AUSDC =
-    (STREAM_AMOUNT_AUSDC / STREAM_DURATION) * STREAM_DURATION;
 
   function execute() external {
     AaveV3Ethereum.COLLECTOR.createStream(
       CHAOS_LABS_TREASURY,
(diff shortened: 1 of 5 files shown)


src/20240415_AaveV3Ethereum_ChaosLabsEngagementAmendment/AaveV3Ethereum_ChaosLabsEngagementAmendment_20240415.sol
```diff
@@ -0,0 +1,26 @@
...
+ * @title Chaos Labs Engagement Amendment
+ * @author Chaos Labs
+ * - Snapshot: https://snapshot.org/#/aave.eth/proposal/0x34b6cf5bc9c8a0525c5b32d4ce2ca2ccfce39d15bd0aba5cab46a4e9e78f3ea8
+ * - Discussion: https://governance.aave.com/t/arfc-chaos-labs-engagement-amendment/17324
+ */
+contract AaveV3Ethereum_ChaosLabsEngagementAmendment_20240415 is IProposalGenericExecutor {
+  address public constant CHAOS_LABS_TREASURY = 0xbC540e0729B732fb14afA240aA5A047aE9ba7dF0;
+  uint256 public constant STREAM_AMOUNT_GHO = 400_000 ether;
+  uint256 public constant STREAM_DURATION = 200 days;
+
```
sakulstra, line 16:
```suggestion
  uint256 public constant ACTUAL_STREAM_AMOUNT =
    (STREAM_AMOUNT_GHO / STREAM_DURATION) * STREAM_DURATION;
    
```

stream amount should always be fully claimable, therefore the amount must be exactly divisible by emissionPerSecond

the final amount will be a few wei less due to rounding: `399999999999999997440000`

src/20240415_AaveV3Ethereum_ChaosLabsEngagementAmendment/AaveV3Ethereum_ChaosLabsEngagementAmendment_20240415.sol
```diff
@@ -0,0 +1,26 @@
...
+  function execute() external {
+    AaveV3Ethereum.COLLECTOR.createStream(
+      CHAOS_LABS_TREASURY,
+      STREAM_AMOUNT_GHO,
```
sakulstra, line 20:
```suggestion
      ACTUAL_STREAM_AMOUNT,
```

sakulstra (changes requested):
Overall it#s looking okay, but it could be better:

1. the amount should always be a multiple of emissionPerSecond to have no unclaimable dust remaining in streams. So what all the other proposals do is sth like: `uint256 exactAmount = (amount/duration) * duration;`

2. the duration on the proposal is a bit arbitrary set to 200 days, which will be slightly longer/shorter then your current streams, but for sure not end at the same time. As you know when your current stream ends you can do it exact actually.

Therefore my suggestion would be to do the following:
```
  uint256 public constant STREAM_AMOUNT_GHO = 400_000 ether;
  uint256 public constant STREAM_END = 1731405179;

  function execute() external {
    uint256 duration = STREAM_END - block.timestamp;
    uint256 exactAmount = (STREAM_AMOUNT_GHO / duration) * duration;
    
    // ... stream creation code
  }
```

With this change the stream will be exact at time of execution & **slightly less** then 400k ether due to rounding(maximum 1 second of emission -1 is lost on the rounding).
Calculated based on current block would be th like `399999999999999982988448`

sakulstra (approved):

rustboyar (approved):

PR #335, opened 2024-05-08
feat: allow to select voting network via generator

brotherlymite:
- [x] Fix js tests

### DIFF
--- a/generator/common.ts
+++ b/generator/common.ts
@@ -1,5 +1,5 @@
 import * as addressBook from '@bgd-labs/aave-address-book';
-import {Options, PoolIdentifier, PoolIdentifierV3, V2_POOLS} from './types';
+import {Options, PoolIdentifier, PoolIdentifierV3, V2_POOLS, VOTING_NETWORK} from './types';
 import {
   arbitrum,
   avalanche,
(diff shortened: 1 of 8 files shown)


generator/common.ts
```diff
@@ -60,6 +60,21 @@ export function getDate() {
...
+  let votingPortal: string;
+
+  if (votingNetwork == VOTING_NETWORK.ETHEREUM) {
+    votingPortal = 'GovernanceV3Ethereum.VOTING_PORTAL_ETH_ETH';
+  } else if (votingNetwork == VOTING_NETWORK.POLYGON) {
+    votingPortal = 'GovernanceV3Ethereum.VOTING_PORTAL_ETH_POL';
+  } else if (votingNetwork == VOTING_NETWORK.AVALANCHE) {
+    votingPortal = 'GovernanceV3Ethereum.VOTING_PORTAL_ETH_AVAX';
+  } else {
+    throw new Error('Invalid voting portal');
```
sakulstra, line 73:
```suggestion
    throw new Error('Invalid voting network');
```

generator/types.ts
```diff
@@ -46,6 +46,7 @@ export interface Options {
   force?: boolean;
   pools: PoolIdentifier[];
   title: string;
+  votingNetwork: VOTING_NETWORK;
```
sakulstra, line 49: perhaps better to make it optional and fallback to now using one?

PR #412, opened 2024-08-01
FIX: May Funding 

defijesus:
update rETH oracle. update LUSD & PYUSD slippages


src/20240603_Multi_MayFundingUpdate/AaveV3Ethereum_MayFundingUpdate_20240603.sol
```diff
@@ -30,7 +30,7 @@ contract AaveV3Ethereum_MayFundingUpdate_20240603 is IProposalGenericExecutor {
   address public constant USDC_FEED = 0x8fFfFfd4AfB6115b954Bd326cbe7B4BA576818f6;
   address public constant DPI_FEED = 0xD2A593BF7594aCE1faD597adb697b5645d5edDB2;
   address public constant USDT_FEED = 0x3E7d1eAB13ad0104d2750B8863b489D65364e32D;
-  address public constant RETH_FEED = 0x536218f9E9Eb48863970252233c8F271f554C2d0;
+  address public constant RETH_FEED = 0x05225Cd708bCa9253789C1374e4337a019e99D56;
```
efecarranza, line 33: This oracle doesn't have a `decimals()` function unfortunately either, so the swaps are not going to go through.
@brotherlymite suggested using a CAPO oracle for this one, as the interface is the same as a Chainlink oracle.
Alternatively (my preferred approach) would be to just withdraw from RocketPool

src/20240603_Multi_MayFundingUpdate/AaveV3Ethereum_MayFundingUpdate_20240603.t.sol
```diff
@@ -283,6 +283,17 @@ contract AaveV3Ethereum_MayFundingUpdate_20240603_Test is ProtocolV3TestBase {
     );
 
     executePayload(vm, address(proposal));
+
+    _baseSwapTest(
```
brotherlymite, line 287: @defijesus we should add this test for all the swaps

brotherlymite:
@defijesus could you also change the [commit hashes](https://github.com/bgd-labs/aave-proposals-v3/blob/main/src/20240603_Multi_MayFundingUpdate/MayFundingUpdate.md?plain=1#L76-L77) to main, so ci can update it to the latest when pr gets merged

brotherlymite (approved):

sendra (approved):

PR #471, opened 2024-10-02
Remove Frax of Isolation Mode - Draft 

marczeller:


marczeller:
Let's wait for risk re-review before moving forward

PR #474, opened 2024-10-03
Update wstETH on Lido

efecarranza:
# Changelog

Update params on Lido instance for wstETH.


efecarranza:
@ianflexa updated! thank you

ianflexa (approved):

PR #494, opened 2024-10-17
fix: set correct asset for CL stream

Rozengarden:


ianflexa:
fixed in https://github.com/bgd-labs/aave-proposals-v3/pull/492

PR #526, opened 2024-11-19
feat: lido gho

sakulstra:
Showcase the idea of using the executor as a facilitator and minting.

### DIFF
--- a/src/20241119_AaveV3EthereumLido_GHOListingOnLidoPool/AaveV3EthereumLido_GHOListingOnLidoPool_20241119.sol
+++ b/src/20241119_AaveV3EthereumLido_GHOListingOnLidoPool/AaveV3EthereumLido_GHOListingOnLidoPool_20241119.sol
@@ -1,14 +1,13 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
+import {AaveV3EthereumAssets} from 'aave-address-book/AaveV3Ethereum.sol';
 import {AaveV3EthereumLido} from 'aave-address-book/AaveV3EthereumLido.sol';
-import {MiscEthereum} from 'aave-address-book/MiscEthereum.sol';
 import {AaveV3PayloadEthereumLido} from 'aave-helpers/src/v3-config-engine/AaveV3PayloadEthereumLido.sol';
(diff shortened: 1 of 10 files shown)


src/20241119_AaveV3EthereumLido_GHOListingOnLidoPool/GHODirectMinter.sol
```diff
@@ -0,0 +1,59 @@
...
+   */
+  function withdrawAndBurn(uint256 amount) external onlyOwner {
+    uint256 amountWithdrawn = POOL.withdraw(address(GHO), amount, address(this));
+
+    GHO.burn(amountWithdrawn);
+  }
+
+  function transferExcessToTreasury() external {
+    (uint256 capacity, ) = GHO.getFacilitatorBucket(address(this));
+    DataTypes.ReserveDataLegacy memory reserveData = POOL.getReserveData(address(GHO));
```
kyzia551, line 54: aToken can be immutable as well, no point to query every time

src/20241119_AaveV3EthereumLido_GHOListingOnLidoPool/AaveV3EthereumLido_GHOListingOnLidoPool_20241119.sol
```diff
@@ -0,0 +1,72 @@
...
+  address public immutable VAULT;
+
+  constructor(address vault) {
+    VAULT = vault;
+  }
+
+  function _postExecute() internal override {
+    IGhoToken(AaveV3EthereumAssets.GHO_UNDERLYING).addFacilitator(
+      VAULT,
+      'LidoD3MVault',
```
kyzia551, line 35:
```suggestion
      'LidoGHOVault',
```

src/20241119_AaveV3EthereumLido_GHOListingOnLidoPool/GHODirectMinter.sol
sakulstra: the reason i query it here is that otherwise i would need to deploy the minter on the proposal, which makes testing a bit madness (as i would need to somehow extract the added address).

If you insist i can do it, but to me feel cleaner this way and 10k gas saved on a rather rare operation probably is not so important. We can still do it in an upgrade.

TepNik:
For me this looks ok.

PR #534, opened 2024-11-25
Aavechan/tbtc wbtc emode

Rozengarden:


marczeller:
We will not escalate this proposal and pause it for now. 

Feel free to drop it from review pipeline.

PR #550, opened 2024-12-20
Claim ZKSync ZK airdrop

MartinGbz:
Instruction to claim ZKSync airdrop are written in this project: https://github.com/ZKsync-Association/zknation-data/blob/main/README.md#l1-smart-contracts-addresses

The [Ethereum Executor](https://etherscan.io/address/0x5300A1a15135EA4dc7aD5a167152C01EFc9b192A) is the address eligible for it.

The process of claim works like this:
- Call a claim function on L1 contract, that will mint the token and send it to the executor address on L2
- Call a transfer function on L1 contract, that will transfer the token from the executor address to a usable address on zksync (here we chose lm.aci.eth => need to confirm).

Each transaction require an amount in ETH to process, so the proposal get the necessary WETH form the collector and unwrap them.

The claim (so the transfer too) is an amount of `8301475000000000000000000`wei ZK. So 8,301,475 ZK. This amount can be found by decoding the generated l2Calldata of the L1 claim function.


MartinGbz:
Note: tried a generation of the calldatas (claim & transfer) 1 week after, and I've still have the same L2 Calldata, so it seems alright.

MartinGbz:
Closing this PR as the Airdrop claim period ended

PR #564, opened 2025-01-09
feat: a.DI Ethereum -> Celo path activation

sendra:


kyzia551 (commented):
Looks fine. But links shoould be added

kyzia551 (approved):

PR #571, opened 2025-01-20
feat: explicitly enable optimizer

### LINKED ISSUE #570: Compiler settings are currently indeterministic
Recent changes to forge build defaults (foundry-rs/foundry#9657) have modified the Solidity optimizer behavior. Prior, optimizer was enabled by default when not turned off explicitly (or runs not provided). Since then, optimizer is now disabled by default. 

Since the optimizer settings are [not explicitly stated right now](https://github.com/bgd-labs/aave-proposals-v3/blob/134c39ccfc2ef4ed515b5d334c6ae59e6320c66e/foundry.toml), it makes builds indeterministic if using an older forge nightly build.
This is especially relevant and breaking since we deterministically deploy payloads using create2 and a different compiler results in a different target address and hence proposal mismatch when trying to verify deployed payload & actions across forge versions.

I would suggest we explicitly specify the optimizer to be enabled with the default (200) runs as it has been prior to this update. This will ensure consistent builds and deployments across all forge versions.

DhairyaSethi:
closes #570

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -3,6 +3,8 @@ src = 'src'
 test = 'tests'
 script = 'scripts'
 solc = '0.8.20'
+optimizer = true
+optimizer_runs = 200
 out = 'out'
 bytecode_hash = 'none'
 libs = ['lib']
@@ -19,6 +21,8 @@ src = 'zksync'
 test = 'zksync'
 libs = ['lib']
 solc = '0.8.20'
+optimizer = true
+optimizer_runs = 200
 fs_permissions = [{ access = "write", path = "./reports" }]
 ffi = true
 evm_version = 'shanghai'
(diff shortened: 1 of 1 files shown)


foundry.toml
```diff
@@ -3,6 +3,8 @@ src = 'src'
 test = 'tests'
 script = 'scripts'
 solc = '0.8.20'
+optimizer = true
+optimizer_runs = 200
```
brotherlymite, line 7: @DhairyaSethi default runs is 200 so maybe we can remove num of runs
```suggestion
```

DhairyaSethi: yeah i explicitly specified in case it's changed by forge (directly, or by moving solar with this change) or by solc in the future xD
since that's very unlikely, lmk if you don't want to indulge the paranoia, will rm

brotherlymite: reference: https://github.com/foundry-rs/foundry/blob/master/crates/config/src/lib.rs#L834

brotherlymite: fine to keep yes

PR #577, opened 2025-01-22
Balancer Claim Mining Rewards

efecarranza:
# Changelog

Allow balancer address to claim mining rewards.

## Note

Where in the tests there is no attempt to claim a reward mentioned in the forum, it's because the rewards have not been initialized yet.


ianflexa (commented):
>Note
Where in the tests there is no attempt to claim a reward mentioned in the forum, it's because the rewards have not been initialized yet.

In this case I think we should add in both tests if the claimer is correctly set in the rewards controller:

```solidity
    assertEq(IRewardsController(AaveV3Ethereum.DEFAULT_INCENTIVES_CONTROLLER).getClaimer(proposal.BALANCER_VAULT()), proposal.CLAIMER());

ianflexa (dismissed):

eboadom:
@efecarranza seems to me there could be a single payload, as the split in one for Ethereum and another for Gnosis at the moment is slightly strange. You define as `immutable` 2 addresses, but actually, you treat them as `constant`, hardcoding.

However, you could just pass them in the constructor as immutable together with the correct emission manager per network, and that way having a single payload

ianflexa (approved):

eboadom (approved):

PR #582, opened 2025-01-24
Remove unnecessary import

efecarranza:
Remove unnecessary import (it's imported in ProtocolBaseTest).


brotherlymite (approved):

brotherlymite:
@efecarranza can you run `npm run test` and then press `u` to update the test snapshots

PR #758, opened 2025-07-03
fix: store cache specific to the evm-version of profile

brotherlymite:
With different evm-version profiles for deploying and testing, when switching from the test profile to deploy profile, we re-compile everything as the evm-version changes and it is a bit pain.

We now store the compiler cache per evm-version, so when switch back and forth between different foundry profiles we don't have to re-compile everything saving us a bit of time


sakulstra (approved):

th30neee (commented):
Nice

PR #772, opened 2025-07-21
feat: add PT sUSDe july and usdtb to pt september emode

Rozengarden:
forum: https://governance.aave.com/t/direct-to-aip-adjustment-of-pt-susde-september-e-mode-and-usdtb-ir-curve/22615
snapshot: direct-to-aip


sakulstra (approved):

brotherlymite:
@Rozengarden can you update the aave-helper submodule to the latest, seems like the build is failing because of that

PR #815, opened 2025-08-29
September Funding Update

efecarranza:
# Changelog

Add September funding update.

## Notes

Voting takes place on Avalanche as requested.
wstETH and wETH are not bridgeable from Optimism to Mainnet with our bridge.
-- Withdrawals will be performed on Monday for tests.

### DIFF
--- a/src/20250826_Multi_SeptemberFundingUpdate/SeptemberFundingUpdate.md
+++ b/src/20250826_Multi_SeptemberFundingUpdate/SeptemberFundingUpdate.md
@@ -2,7 +2,6 @@
 title: "September Funding Update"
 author: "@TokenLogic"
 discussions: "https://governance.aave.com/t/direct-to-aip-september-2025-funding-update/22915"
-snapshot: "Direct-to-AIP"
 ---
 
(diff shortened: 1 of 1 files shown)


src/20250826_Multi_SeptemberFundingUpdate/AaveV3Ethereum_SeptemberFundingUpdate_20250826.sol
```diff
@@ -0,0 +1,142 @@
...
+      MERIT_GHO_ALLOWANCE
+    );
+  }
+
+  function _swaps() internal {
+    // ETH
+    AaveV3Ethereum.COLLECTOR.transfer(
+      IERC20(AaveV3Ethereum.COLLECTOR.ETH_MOCK_ADDRESS()),
+      address(this),
+      ETH_TO_DEPOSIT
```
sakulstra, line 96: i think fine, but perhaps in the future better use .balance instead of fixed amount.

sakulstra (approved):

PR #824, opened 2025-09-11
RewardsSteward + SwapSteward

efecarranza:
# Changelog

Add RewardsSteward across multiple chains.
Add MainnetSwapSteward on Mainnet.


src/20250821_Multi_StewardDeploymentMainnetSwapStewardAndRewardsSteward/AaveV3Ethereum_StewardDeploymentMainnetSwapStewardAndRewardsSteward_20250821.sol
```diff
@@ -0,0 +1,252 @@
...
+    // Token: USDT
+    _setOracleAndBudget(
+      AaveV3EthereumAssets.USDT_UNDERLYING,
+      AaveV3EthereumAssets.USDT_ORACLE,
+      USD_STABLE_BUDGET * 10 ** AaveV3EthereumAssets.USDT_DECIMALS
+    );
+    // Token: USDe
+    _setOracleAndBudget(
+      AaveV3EthereumAssets.USDe_UNDERLYING,
+      AaveV3EthereumAssets.USDe_ORACLE,
```
pavelvm5, line 78: We are using capped `USDT/USD` oracle for `USDe`, so I think here we should replace it with real `USDe` feed.

Also, I'm not so sure, that we should use `capped` versions of oracles for swaps. I guess better to switch to normal ones. For almost all stablecoins they could be received from `oracle.ASSET_TO_USD_AGGREGATOR()`.

src/20250821_Multi_StewardDeploymentMainnetSwapStewardAndRewardsSteward/AaveV3Ethereum_StewardDeploymentMainnetSwapStewardAndRewardsSteward_20250821.sol
```diff
@@ -0,0 +1,252 @@
...
+  uint256 public constant RLUSD_BUDGET = 1_000_000 ether;
+  uint256 public constant LUSD_BUDGET = 500_000 ether;
+  uint256 public constant FRAX_BUDGET = 150_000 ether;
+  uint256 public constant PYUSD_BUDGET = 20_000e6;
+  uint256 public constant UNI_BUDGET = 15_000 ether;
+  uint256 public constant MKR_BUDGET = 100 ether;
+  uint256 public constant ONEINCH_BUDGET = 50_000 ether;
+  uint256 public constant ENS_BUDGET = 200 ether;
+  uint256 public constant SNX_BUDGET = 150_000 ether;
+
```
pavelvm5, line 41: On the forum there're also `SPK` and `SAFE`? Why aren't they here?

pavelvm5 (dismissed):

brotherlymite (approved):

pavelvm5 (approved):

PR #1017, opened 2026-03-10
Reduce safety module emissions

Joafai:


pavelvm5 (changes requested):
Could you also merge main, cause tests should be fixed for now.

pavelvm5 (approved):

sakulstra (approved):

PR #1075, opened 2026-05-01
fix : generator improvements

Kogaroshi:
- allow to select new asset for eModes during listing in prompts
- more consistency on Explorer URL comments
- add Price Feed as a constant when adding new assets (instead of inlined in execution)

### DIFF
--- a/generator/features/assetListing.spec.ts
+++ b/generator/features/assetListing.spec.ts
@@ -12,6 +12,7 @@ describe('feature: assetListing', () => {
       pool: 'AaveV3Ethereum',
       cfg: assetListingConfig,
       cache: {blockNumber: 42},
+      configs: {},
     });
     expect(output).toMatchSnapshot();
   });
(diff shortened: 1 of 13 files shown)


generator/features/assetListing.spec.ts
```diff
@@ -12,6 +12,7 @@ describe('feature: assetListing', () => {
       pool: 'AaveV3Ethereum',
       cfg: assetListingConfig,
       cache: {blockNumber: 42},
+      configs: {},
```
DhairyaSethi, line 15: can you add votingNetwork on this config as well, generator generated it but it always fails type check

generator/features/assetListing.ts
```diff
@@ -64,7 +64,7 @@ async function fetchCustomImpl(): Promise<TokenImplementations> {
 function generateAssetListingSol(cfg: Listing) {
   return `asset: ${cfg.assetSymbol},
   assetSymbol: "${cfg.assetSymbol}",
-  priceFeed: ${translateJsAddressToSol(cfg.priceFeed)},
+  priceFeed: ${cfg.assetSymbol}_PRICE_FEED,
```
DhairyaSethi, line 67: how does this help? will it work for new asset?

DhairyaSethi (approved):
at your discretion

generator/features/assetListing.ts
Kogaroshi: yes, for new assets it will work as before, just moving the inlined price feed address given during listing to a constant, added later via `listingConstant += `address public constant ${cfg.assetSymbol}_PRICE_FEED = ${translateJsAddressToSol(cfg.priceFeed)};\n`;`

PR #1083, opened 2026-05-05
SGho Activation

efecarranza:
The following checklist is only relevant for proposal PRs. Remove this section if not applicable.

### Pre-review checklist:

- [ ] If deploy scripts are manually updated from the generated ones, I have carefully validated that they are correct, including the deploy commands and the proposal-creation script.


src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.sol
```diff
@@ -0,0 +1,71 @@
...
+interface IsGhoSteward {
+  struct RateConfig {
+    uint16 amplification;
+    uint16 floatRate;
+    uint16 fixedRate;
+  }
+
+  function setRateConfig(RateConfig calldata rateConfig) external returns (uint16);
+  function setSupplyCap(uint256 supplyCap) external;
+}
```
Kogaroshi: Please move to `src/interfaces`, we might need it again in the future.

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.sol
```diff
@@ -0,0 +1,71 @@
...
+  bytes32 public constant TOKEN_RESCUER_ROLE = keccak256('TOKEN_RESCUER_ROLE');
+  bytes32 public constant YIELD_MANAGER_ROLE = keccak256('YIELD_MANAGER_ROLE');
+
+  uint256 public constant GHO_ALLOWANCE = 10_000_000 ether;
+  uint256 public constant SUPPLY_CAP = 400_000_000 ether;
+  uint16 public constant FIXED_RATE = 4_25;
+
+  function execute() external {
+    IAccessControl(SGHO_STEWARD).grantRole(FIXED_RATE_MANAGER_ROLE, address(this));
+    IAccessControl(SGHO_STEWARD).grantRole(SUPPLY_CAP_MANAGER_ROLE, address(this));
```
Kogaroshi: either use `GovernanceV3Ethereum.EXECUTOR_LVL_1` like for granting `TOKEN_RESCUER_ROLE`, but please make them match.

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.t.sol
```diff
@@ -0,0 +1,189 @@
...
+        proposal.FIXED_RATE_MANAGER_ROLE(),
+        GovernanceV3Ethereum.EXECUTOR_LVL_1
+      )
+    );
+    assertFalse(
+      IAccessControl(proposal.SGHO_STEWARD()).hasRole(
+        proposal.SUPPLY_CAP_MANAGER_ROLE(),
+        GovernanceV3Ethereum.EXECUTOR_LVL_1
+      )
+    );
```
Kogaroshi: what entities will have those Roles post execution of the AIP ?

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.sol
```diff
@@ -0,0 +1,71 @@
...
+  bytes32 public constant SUPPLY_CAP_MANAGER_ROLE = keccak256('SUPPLY_CAP_MANAGER_ROLE');
+
+  // sGho roles
+  bytes32 public constant PAUSE_GUARDIAN_ROLE = keccak256('PAUSE_GUARDIAN_ROLE');
+  bytes32 public constant TOKEN_RESCUER_ROLE = keccak256('TOKEN_RESCUER_ROLE');
+  bytes32 public constant YIELD_MANAGER_ROLE = keccak256('YIELD_MANAGER_ROLE');
+
+  uint256 public constant GHO_ALLOWANCE = 10_000_000 ether;
+  uint256 public constant SUPPLY_CAP = 400_000_000 ether;
+  uint16 public constant FIXED_RATE = 4_25;
```
Kogaroshi: pls add tests to verify those Roles constants against contract getters, and to confirm uint constants are correct.

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.sol
avniculae: I'm personally fine with using `address(this)` here. I get it's the same as EXECUTOR_LVL_1, but conceptually makes sense to use `address(this)` for this temporary permissions imo, wdyt?

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.sol
avniculae: +1, we can even paste entire interface from gho-origin

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.t.sol
```diff
@@ -0,0 +1,189 @@
...
+    uint16 amplification;
+    uint16 floatRate;
+    uint16 fixedRate;
+  }
+
+  function getRateConfig() external view returns (RateConfig memory);
+  function sGHO() external view returns (address);
+}
+
+interface ISGho is IERC4626 {
```
avniculae: same applies here, interfaces should be moved to `src/interfaces/` and imported from there

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.t.sol
```diff
@@ -0,0 +1,189 @@
...
+      proposal.GHO_ALLOWANCE()
+    );
+
+    vm.startPrank(MiscEthereum.AFC_SAFE);
+    IERC20(GhoEthereum.GHO_TOKEN).transferFrom(
+      address(AaveV3Ethereum.COLLECTOR),
+      proposal.SGHO(),
+      100_000 ether
+    );
+    vm.stopPrank();
```
avniculae, line 299: let's check balance of AFC_SAFE before and after

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.sol
Kogaroshi: yes, my comment was somehow incomplete, I just want to have consistency in Payload, either using `address(this)` everywhere, or the address-book ref everywhere.

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.t.sol
```diff
@@ -0,0 +1,189 @@
...
+  }
+}
```
avniculae, line 306: I think we should have a test to make sure roles in SGHO_STEWARD are awarded to risk council (these are already true given deployment, but best to check it here as well)

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.sol
avniculae: I am fine with this inconsistency, bc conceptually they are "different". EXECUTOR_LVL_1 is used as per the gov post, and `address(this)` is used for these temp roles, wdyt?

Kogaroshi: I understand the point of having `EXECUTOR_LVL_1` as specified in the AIP text, but they are not "different", in reality it's about granting role to a given entity, and only later to remove those granted roles.
If the "difference" is because of how it is presented in the AIP text, maybe the Specification part should be more detailed to explain how the `Supply Cap` and `Target Rate` (be detailing how the specific roles are given to the Executor so those calues can be set as part of the Payload).
Even if the role is granted just in the context of the Payload, we are still granting a critical role, and it should be done and documented with the same rigor as any other Role management action.

avniculae: thanks, let's address then @efecarranza

efecarranza: @Kogaroshi @avniculae 
yeah, they are the same. i did it this way just to highlight that one was temporary vs. permanent.
in reality, there's no real reason to even remove them, since governance is fine to have these controls.
what do you think about not even removing? we can add to the proposal and .md file text that they are permanent.

Kogaroshi: I'm also okay with not removing the role at the end of the Payload, but in that case the attribution of roles must be specified in the AIP-text, and matching the other role attributions (no `address(this)`)

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.t.sol
efecarranza: added test to show risk council has the roles

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.t.sol
```diff
@@ -50,10 +35,10 @@ contract AaveV3Ethereum_SGhoLaunch_20260427_Test is ProtocolV3TestBase {
   }
 
   function test_initialization() public {
-    assertEq(ISGho(proposal.SGHO()).targetRate(), 0);
-    assertEq(ISGho(proposal.SGHO()).supplyCap(), 0);
-    assertEq(ISGho(proposal.SGHO()).GHO(), GhoEthereum.GHO_TOKEN);
-    assertEq(ISGho(proposal.SGHO()).asset(), GhoEthereum.GHO_TOKEN);
+    assertEq(IsGho(proposal.SGHO()).targetRate(), 0);
```
Kogaroshi: `IsGho(proposal.SGHO()` can be set as a private var during `setUp` to simplify a bit the tests.

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.sol
avniculae: tagging @DhairyaSethi here for his input

src/20260427_AaveV3Ethereum_SGhoLaunch/SGhoLaunch.md
```diff
@@ -40,9 +40,12 @@ Market opportunity. With the Sky Savings Rate at 3.75% APR and broader DeFi yiel
...
 
-Grant the PAUSE_GUARDIAN_ROLE to the PROTOCOL_GUARDIAN SAFE.
-Grant the TOKEN_RESCUER_ROLE to the Governance LVL 1 Executor.
-Grant the YIELD_MANAGER_ROLE to the sGhoSteward contract.
+Grant the sGho PAUSE_GUARDIAN_ROLE to the PROTOCOL_GUARDIAN SAFE.
+Grant the sGho YIELD_MANAGER_ROLE to the sGhoSteward contract.
+
+Grant the sGhoSteward FIXED_RATE_MANAGER_ROLE to the Governance LVL 1 Executor.
+Grant the sGhoSteward SUPPLY_CAP_MANAGER_ROLE to the Governance LVL 1 Executor.
+Grant the sGho TOKEN_RESCUER_ROLE to the Governance LVL 1 Executor.
```
avniculae, line 48: maybe we can move this back up, in the "section" of sGho roles? nbd

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.t.sol
```diff
@@ -0,0 +1,232 @@
...
+    assertTrue(
+      IAccessControl(proposal.SGHO_STEWARD()).hasRole(
+        sghoSteward.SUPPLY_CAP_MANAGER_ROLE(),
+        GovernanceV3Ethereum.EXECUTOR_LVL_1
+      )
+    );
+  }
+
+  function test_rolesSGhoSteward() public {
+    // RiskCouncil already has these roles from deployment
```
avniculae, line 206: on the same note, let's check executor lvl1 has DEFAULT_ADMIN_ROLE before and after proposal execution

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.sol
DhairyaSethi: i would explicitly always only grant to executor from address book and not address(this) (even though they are the same here) and yep good with leaving the roles on executor & being explicit about it in the text

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.sol
```diff
@@ -0,0 +1,63 @@
...
+      IsGho(SGHO).PAUSE_GUARDIAN_ROLE(),
+      MiscEthereum.PROTOCOL_GUARDIAN
+    );
+    IAccessControl(SGHO).grantRole(
+      IsGho(SGHO).TOKEN_RESCUER_ROLE(),
+      GovernanceV3Ethereum.EXECUTOR_LVL_1
+    );
+    IAccessControl(SGHO).grantRole(IsGho(SGHO).YIELD_MANAGER_ROLE(), SGHO_STEWARD);
+
+    IsGhoSteward(SGHO_STEWARD).setSupplyCap(SUPPLY_CAP);
```
avniculae, line 52: I'm realising this deviates from how supply caps work in v3 and v4, where caps are set without the token decimals. We need to be very careful with this, can you pls add a test to make sure sgho reverts if we try to supply SUPPLY_CAP + 1 tokens?

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.t.sol
```diff
@@ -0,0 +1,232 @@
...
+        MiscEthereum.AFC_SAFE
+      ),
+      proposal.GHO_ALLOWANCE()
+    );
+
+    uint256 balanceBefore = IERC20(GhoEthereum.GHO_TOKEN).balanceOf(proposal.SGHO());
+    uint256 transferAmount = 100_000 ether;
+
+    vm.startPrank(MiscEthereum.AFC_SAFE);
+    IERC20(GhoEthereum.GHO_TOKEN).transferFrom(
```
avniculae, line 294: let's have a test that checks allowance as well. we can check that allowance (10m) works, 10m + 1 reverts

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.sol
```diff
@@ -0,0 +1,63 @@
...
+contract AaveV3Ethereum_SGhoLaunch_20260427 is IProposalGenericExecutor {
+  // https://etherscan.io/address/0xE1753F2e00940cC31213dd92013cF019DFE4ca1d
+  address public constant SGHO = 0xE1753F2e00940cC31213dd92013cF019DFE4ca1d;
+
+  // https://etherscan.io/address/0x60Bf2DF49F17529Cf956D57848ebEB8a0d0a2757
+  address public constant SGHO_STEWARD = 0x60Bf2DF49F17529Cf956D57848ebEB8a0d0a2757;
+
+  uint256 public constant GHO_ALLOWANCE = 10_000_000 ether;
+  uint256 public constant SUPPLY_CAP = 400_000_000 ether;
+  uint16 public constant FIXED_RATE = 4_25;
```
avniculae, line 30: can we add a test that supplies to sgho, advances time and verifies that we accrued 4.25%? just to ensure format for setting this is correct

src/20260427_AaveV3Ethereum_SGhoLaunch/SGhoLaunch.md
```diff
@@ -0,0 +1,59 @@
...
+author: "TokenLogic"
+discussions: "https://governance.aave.com/t/arfc-sgho-launch-configuration/24346"
+snapshot: "https://snapshot.box/#/s:aavedao.eth/proposal/0xb9e9b01efcf6151bade78546d0f51f11d7961939b649fb7717e82ea3d43d4f47"
+---
+
+## Simple Summary
+
+This publication provides an overview of the new native on-chain yield-accruing sGHO savings product upgrade. Within, we define the initial parameter configuration, present the migration process from the current sGHO (stkGHO rebranded), and introduce the new sGHO admin steward role.
+
+The proposal sets a fixed Aave Savings Rate (ASR) of `4.25% APR` (50 basis points above the Sky Savings Rate), refines the boost program from 8 boosts to 1 transitional boost, establishes the sGHO Steward as the rate governance mechanism, and defines an aggressive migration schedule from the legacy sGHO contract.
```
avniculae, line 12: Some parts are not applicable for this proposal per se. I think we could remove some of the parts, eg "defines an aggressive migration schedule from the legacy sGHO contract.", "(50 basis points above the Sky Savings Rate),". This note applies to other parts of this write up as well. Maybe we can say that this proposal performs the initial tasks required for achieving these final goals. worth changing or you think we should keep as is? cc @Kogaroshi

Kogaroshi: Those comments might come from previous governance steps, and while they are not required for the AIP specifications itself, it can be okay to give them to extend the context.

src/20260427_AaveV3Ethereum_SGhoLaunch/SGhoLaunch.md
efecarranza: i think listed here in the specification section of what the proposal is doing is a bit better in this case

avniculae: I meant 3 rows above, but nbd

src/20260427_AaveV3Ethereum_SGhoLaunch/AaveV3Ethereum_SGhoLaunch_20260427.sol
efecarranza: added test, but yeah, supply cap here is in GHO units:
https://github.com/aave-dao/gho-origin/blob/main/docs/sgho-precision-analysis/precision.md#5-supply-cap

avniculae (approved):

Kogaroshi (approved):

PR #1086, opened 2026-05-06
Extend stkaave emissions

Joafai:
### Pre-review checklist:

- [ ] If deploy scripts are manually updated from the generated ones, I have carefully validated that they are correct, including the deploy commands and the proposal-creation script.


meng-xu-cs:
Hi, apologies for a direct comment on the PR.

While the PR seems find for me, the linked Snapshot/forum can imply that `stkAAVE` emissions are an active governance topic, but they do not imply this specific payload action: an Ecosystem Reserve AAVE allowance top-up for stkAAVE using `currentAllowance + emissionPerSecond * 90 days`.

I am wondering did I miss any of the discussion that this should be the way to extend the stream or am I just being blind here?

Joafai:
hi @meng-xu-cs  , thanks for checking!
what we've been doing in the past is add a comment and reference that in the PR, we just added the comment here: https://governance.aave.com/t/arfc-amend-safety-module-emissions/16640/52
for reference here's also a previous PR where we do the same: https://github.com/aave-dao/aave-proposals-v3/pull/961/changes#diff-99c86ed766f5788fde1c0507724bdee57da2a1463dd5b2540b47e8133813078c

we've done multiple instances of these in the past. please let me know if that comment is sufficient or we can do it a bit differently. thank you

src/20260505_AaveV3Ethereum_ExtendStkAAVEEmissions/AaveV3Ethereum_ExtendStkAAVEEmissions_20260505.sol
```diff
@@ -0,0 +1,44 @@
...
+    uint256 newAllowance = currentAllowance + (uint256(emissionPerSecond) * 90 days);
+
+    MiscEthereum.AAVE_ECOSYSTEM_RESERVE_CONTROLLER.approve(
+      MiscEthereum.ECOSYSTEM_RESERVE,
+      AaveV3EthereumAssets.AAVE_UNDERLYING,
+      AaveSafetyModule.STK_AAVE,
+      0
+    );
+
+    MiscEthereum.AAVE_ECOSYSTEM_RESERVE_CONTROLLER.approve(
```
Kogaroshi, line 37: could use forceApprove here as well

meng-xu-cs (approved):

Kogaroshi (approved):

PR #1094, opened 2026-05-13
[Skyward] Chaos offboarding part 3

MartinGbz:
### Pre-review checklist:

- [x] If deploy scripts are manually updated from the generated ones, I have carefully validated that they are correct, including the deploy commands and the proposal-creation script.
- [x] If the aave-helpers submodule is updated, I have validated it is pointing to the latest version.
- [x] For an asset listing, the write-up includes a detailed specification of the price feed used, CAPO adapters (with each CAPO layer described separately), and eModes (with tables) if changed.


src/20260507_AaveV3EthereumLido_OffboardingPlanForChaosLabsPart3CancelStream/config.ts
```diff
@@ -0,0 +1,15 @@
+import {ConfigFile} from '../../generator/types';
+export const config: ConfigFile = {
+  rootOptions: {
+    author: 'ChaosLabs (performed by Skyward)',
+    pools: ['AaveV3EthereumLido'],
+    title: 'Orderly Transition and Offboarding Plan for Chaos Labs part2',
```
Kogaroshi: might need a title change, now that it is "part 3", and just cancels the stream

src/20260507_AaveV3EthereumLido_OffboardingPlanForChaosLabsPart3CancelStream/config.ts
```diff
@@ -0,0 +1,15 @@
...
+    shortName: 'OrderlyTransitionAndOffboardingPlanForChaosLabsPart2',
```
Kogaroshi: those 2 were not changed/reverted to "part 2"

Kogaroshi (approved):

PR #1134, opened 2026-06-23
fix: remove generator whitelabel blocker

Kogaroshi:
Remove blockers preventing to have Whitelabel payloads with other payloads


generator/templates/aip.template.ts
```diff
@@ -32,6 +33,7 @@ ${Object.keys(configs)
 ## References
 
 - Implementation: ${options.markets
+    .filter((market) => !isWhitelabelMarket(market))
```
DhairyaSethi, line 36: why skip for whitelabel?

Kogaroshi: whitelabel payloads shouldn't be in the AIP itself imo

generator/templates/script.template.ts
```diff
@@ -44,6 +43,9 @@ export function generateScript(options: Options) {
   template += Object.keys(marketsToChainsMap)
     .filter((c) => c !== 'ZkSync')
     .map((chain) => {
+      const groupIsWhitelabel = marketsToChainsMap[chain].some(({market}) =>
```
avniculae: `.some()` here means a chain with both a DAO and a whitelabel market gets marked permissioned, and the DAO payload silently drops out of the proposal. Guarding it properly means splitting the chain group by whitelabel-ness — tightening `isWhitelabelMarket` alone won't do it, since both markets still map to the same chain.

avniculae:
`generator/README.md:53` still says whitelabel and regular markets can't be mixed — no longer true after this PR, should be removed.

generator/templates/script.template.ts
Kogaroshi: good catch.
I think I will change to instead have distinct payloads between owned and whitelabel ones on the same chain, as it may occur in the future, and some updates (to Price Adapters or Stewards) might need to impact all instances of the same chain at once.

Kogaroshi: 6965201918813d64cfedbb06d996e10602b67a31

avniculae:
`generator/utils/importsResolver.ts:134` hardcodes the import for `GovernanceV3InkWhitelabel`, but `script.template.ts` now generates `GovernanceV3${suffix}.PERMISSIONED_PAYLOADS_CONTROLLER` for any whitelabel market. A second whitelabel market would produce a script referencing a symbol that never gets imported, so it won't compile. The `GovernanceV3Ethereum` check just above also matches as a substring, so an `EthereumWhitelabel` would pull in the wrong lib. Could we derive this import the same generic way the template does instead of one line per market?

avniculae (approved):

PR #1152, opened 2026-07-08
feat: expand generated tests proposal coverage

AlbertoCentonze:
Expands the proposal generator test output to cover v3 reserve configuration expectations more explicitly.

This adds generated test helpers for:
- asset listings and custom listings
- borrow updates
- cap updates
- collateral updates
- freeze/unfreeze updates
- untouched reserves do not change

It also adds generator-level coverage and snapshots for the new emitted test code.


.gitmodules
```diff
@@ -3,4 +3,5 @@
 	url = https://github.com/aave-dao/aave-umbrella
 [submodule "lib/aave-helpers"]
 	path = lib/aave-helpers
-	url = https://github.com/aave-dao/aave-helpers
+	url = https://github.com/aave-dao/aave-helpers.git
+	branch = feat/v3-reserve-utils
```
AlbertoCentonze: This needs to go away once https://github.com/aave-dao/aave-helpers/pull/761 is merged

generator/templates/test.template.ts
```diff
@@ -32,7 +32,29 @@ export const testTemplate = (
 
   const defaultTestCall = v4
     ? `defaultTest('${contractName}', address(proposal));`
-    : `defaultTest('${contractName}', ${market}.POOL, address(proposal) ${isWhitelabelMarket(market) ? ', true, true' : ''});`;
+    : `defaultTest('${contractName}', ${market}.POOL, address(proposal)${isWhitelabelMarket(market) ? ', true, true' : ''});`;
```
Kogaroshi: linting removed the space here ?

Kogaroshi (dismissed):
lgtm, pls fix conflicts

.gitmodules
Kogaroshi: you can revert back to main now, and make sure to update foundry.lock

generator/features/collateralsUpdates.ts
```diff
@@ -1,5 +1,4 @@
 import {CodeArtifact, FEATURE, FeatureModule, MarketIdentifier} from '../types';
-import {percentInput} from '../prompts';
```
Kogaroshi, line 2: this was unused ?

AlbertoCentonze: Yep

AlbertoCentonze: The actual prompting is done through percentPrompt, inside `fetchCollateralUpdate` with `percentPrompt`

Kogaroshi (dismissed):

generator/features/collateralsUpdates.ts
AlbertoCentonze: https://github.com/aave-dao/aave-proposals-v3/pull/1155 loosely related cleanup found when I was checking for this

generator/templates/test.template.ts
```diff
@@ -25,14 +25,49 @@ export const testTemplate = (
...
+
+  const updatedAssets = Array.from(
+    new Set(
+      marketConfig.artifacts
+        .map((artifact) => artifact.test?.updatedAssets)
+        .flat()
+        .filter((asset) => asset !== undefined),
+    ),
+  );
+  const reserveConfigChangesTest =
```
avniculae, line 58: This breaks generated ZkSync tests: `AaveV3ZkSync` also resolves to `testBase === 'ProtocolV3TestBase'`, but the import points at `aave-helpers/zksync/src/ProtocolV3TestBase.sol`, which is a separate contract without `reserveConfigChangesTest`, the `_expected*` virtual hooks, or the `ExpectedListing` export. Reproduced locally: rendered the template for `AaveV3ZkSync` and compiled the emitted body against the zksync test base — fails with `Undeclared identifier: reserveConfigChangesTest` (and the `_expected*` overrides have nothing to override). Either port the new helpers to the zksync test base in aave-helpers, or gate on `chain !== 'ZkSync'` — in which case the feature `build()` fns need the same gating for their `_expected*` test fns, not just the template.

generator/features/assetListing.ts
```diff
@@ -229,6 +272,13 @@ export const assetListingCustom: FeatureModule<ListingWithCustomImpl[]> = {
     return response;
   },
   build({market, cfg}) {
+    const listingTests = cfg.map(
+      (cfg) => `function test_dustBinHas${cfg.base.assetSymbol}Funds() public {
+            ${testExecuteProposal(market)}
+            address aTokenAddress = ${market}.POOL.getReserveAToken(proposal.${cfg.base.assetSymbol}());
+            assertGte(IERC20(aTokenAddress).balanceOf(${market}.DUST_BIN), 10 ** ${cfg.base.decimals});
```
avniculae: Since this code is being moved anyway: `assertGte` isn't a forge-std function (it's `assertGe`), so the custom-listing dust bin test has never compiled. Worth fixing here.

generator/templates/test.template.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+    expect(output).toContain("'AaveV3Ethereum_Test_20231023'");
+    expect(output).toContain('AaveV3Ethereum.POOL');
+    expect(output).toContain('address(proposal)');
+    expect(output).not.toContain('defaultTestWithReserveConfigChanges');
+    expect(output).not.toContain('function _payload()');
+    expect(output).not.toContain('function _pool()');
+    expect(output).not.toContain('_allReserveChangesDeclared');
+  });
+
+  it('emits an unchanged-reserve-config test for v3 payloads without expected changes', () => {
```
avniculae, line 71: This doesn't cover what the name says: `marketConfig()` always injects an artifact with one entry in `updatedAssets`, so this still asserts `new address[](1)`. To actually test the no-changes path, pass an artifact without test data and assert `new address[](0)`.

generator/features/assetListing.ts
```diff
@@ -86,6 +86,48 @@ function generateAssetListingSol(cfg: Listing) {
   })`;
 }
 
+function generateExpectedListingSol(cfg: Listing) {
```
avniculae: `generateExpectedListingSol` duplicates `generateAssetListingSol` except for the `asset`/`priceFeed` lines. Parameterizing how those two fields render would keep the field lists from drifting apart.

generator/templates/test.template.ts
avniculae: zk sync is deprecated and has 1M in TVL, but not sure if we should deprecate support in this repo yet. wdyt

Kogaroshi: zkSync tooling should be deprecated tho, is it worth to try and fix it there ?

generator/templates/test.template.spec.ts
avniculae: can you explain what's the intention with this test?

generator/templates/test.template.ts
AlbertoCentonze: FWIW zksync generated tests were already broken before this PR. I just left any zksync logic as is because repairing tests there would make this PR scope creep quite a lot. 

An example of divergence that I spotted with the help of codex:
> There is a zkSync helper mismatch coming from `lib/aave-helpers` being checked out at a newer commit than the parent repo pins. That newer helper code serializes fields like `ReserveConfig.debtCeiling`, `isSiloed`, and `isBorrowableInIsolation`, but the resolved `aave-v3-origin-tests` `ReserveConfig` struct does not contain those fields, so compilation fails once the zkSync test path is built.

I'm happy to try fixing this in a future PR if we all deem it worth it.

generator/features/assetListing.ts
AlbertoCentonze: Good catch, since I didn't have to work on a real listing this slipped through. I'll make sure this works e2e.

generator/templates/test.template.spec.ts
AlbertoCentonze: Making sure that when there is no change in the reserves `test_reserveConfigChanges` still gets generated. The presence of that test case will check that all reserves are indeed unchanged. Can have less assertions if it makes the intent of the test clearer.

AlbertoCentonze: Meanwhile [description matches test](https://github.com/aave-dao/aave-proposals-v3/pull/1152/commits/1f3123f098d6cda05582a87df778b796623b62e5)

generator/features/assetListing.ts
AlbertoCentonze: Fixed in https://github.com/aave-dao/aave-proposals-v3/pull/1152/changes/faecdec38dc297f6996b50f14dc57a486721c02a

Not super happy with the typescript wizardry but that's the cleanest approach I could think of without messing too much with types or increasing change surface. Let me know if it's ok.

generator/features/assetListing.ts
AlbertoCentonze: PR scope creeped a bit because of this but now there's a lot of unit tests that actually compile the generated code. 

https://github.com/aave-dao/aave-proposals-v3/blob/630b28eb4a5cff37e96493f4b8ac13e7936f54d5/generator/features/assetListing.spec.ts#L62-L92

AlbertoCentonze: These tests are kind of e2e not unitary anymore, should I separate them better perhaps?

.github/workflows/test.yml
```diff
@@ -55,4 +55,17 @@ jobs:
           exit 1
 
   test-js:
-    uses: aave-dao/github-workflows/.github/workflows/test-node.yml@main
```
AlbertoCentonze, line 58: can add a foundry flag to test-node if we're not happy with the handrolled workflow here

PR #1164, opened 2026-07-16
July 2026 Funding Update

efecarranza:
### Pre-review checklist:

- [x] If deploy scripts are manually updated from the generated ones, I have carefully validated that they are correct, including the deploy commands and the proposal-creation script.


efecarranza:
@avniculae has a commented code that i'll remove if it can't be done but a user sent 25,000 USDC to aEthUSDC directly 😬 been looking at a way to recover but I don't it's possible with the tools available?

src/20260715_Multi_July2026FundingUpdate/AaveV3Arbitrum_July2026FundingUpdate_20260715.sol
```diff
@@ -0,0 +1,87 @@
...
+contract AaveV3Arbitrum_July2026FundingUpdate_20260715 is IProposalGenericExecutor {
+  uint256 public constant WETH_ALLOWANCE = 160 ether;
+  uint256 public constant USDCn_ALLOWANCE = 150_000e6;
+  uint256 public constant USDT_ALLOWANCE = 50_000e6;
+  uint256 public constant GHO_ALLOWANCE = 55_000 ether;
+
+  function execute() external {
+    _depositEth();
+    _allowances();
+    _cancelOldAllowances();
```
Kogaroshi: can we cancel old ones before setting new allowances pls

src/20260715_Multi_July2026FundingUpdate/AaveV3Arbitrum_July2026FundingUpdate_20260715.t.sol
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  function test_approvals_weth() public {
+    assertGt(
+      IERC20(AaveV3ArbitrumAssets.WETH_A_TOKEN).allowance(
+        address(AaveV3Arbitrum.COLLECTOR),
+        MiscArbitrum.AFC_SAFE
+      ),
+      0
+    );
```
Kogaroshi: is this just to ensure the allowance is non 0 ? I don't see any other utility given the next assert. If so, please modify it to be more explicit

src/20260715_Multi_July2026FundingUpdate/AaveV3Base_July2026FundingUpdate_20260715.sol
```diff
@@ -0,0 +1,29 @@
...
+      IERC20(AaveV3Base.COLLECTOR.ETH_MOCK_ADDRESS()),
+      address(this),
+      collectorEthBalance
+    );
+    IWrappedTokenGatewayV3(AaveV3Base.WETH_GATEWAY).depositETH{value: collectorEthBalance}(
+      address(AaveV3Base.POOL),
+      address(AaveV3Base.COLLECTOR),
+      0
+    );
+  }
```
Kogaroshi, line 28: it seems that deposit raw ETH action is done on all networks, please make it a common base function inherited by all payload to ensure consistency

Kogaroshi: and can be the same for the test for that action

src/20260715_Multi_July2026FundingUpdate/AaveV3Ethereum_July2026FundingUpdate_20260715.sol
```diff
@@ -0,0 +1,194 @@
...
+
+  function _cancelAllowances() internal {
+    AaveV3Ethereum.COLLECTOR.approve(
+      IERC20(AaveV3EthereumAssets.USDC_A_TOKEN),
+      MiscEthereum.MERIT_AHAB_SAFE,
+      0
+    );
+  }
+
+  function _monad() internal {
```
Kogaroshi: can we rename to give more context here ?

src/20260715_Multi_July2026FundingUpdate/AaveV3Ethereum_July2026FundingUpdate_20260715.t.sol
```diff
@@ -0,0 +1,332 @@
...
+
+    uint256 allowanceAfter = IERC20(AaveV3EthereumLidoAssets.GHO_A_TOKEN).allowance(
+      address(AaveV3Ethereum.COLLECTOR),
+      proposal.TOKEN_LOGIC()
+    );
+
+    assertEq(allowanceAfter, allowanceBefore + proposal.REIMBURSEMENTS_GHO_AMOUNT());
+  }
+
+  function test_replenishSwapTokenBudget() public {
```
Kogaroshi, line 181: maybe break down this test to have an internal helper, and loop over the list of assets, it will be easier to read

src/20260715_Multi_July2026FundingUpdate/AaveV3Plasma_July2026FundingUpdate_20260715.sol
```diff
@@ -0,0 +1,103 @@
...
+    );
+
+    AaveV3Plasma.COLLECTOR.approve(
+      IERC20(AaveV3PlasmaAssets.sUSDe_A_TOKEN),
+      MiscPlasma.AFC_SAFE,
+      IERC20(AaveV3PlasmaAssets.sUSDe_A_TOKEN).balanceOf(address(AaveV3Plasma.COLLECTOR))
+    );
+  }
+
+  function _refreshUsdtAllowance() internal {
```
Kogaroshi: one allowance here is canceled, the "refresh" name is a bit misleading

src/20260715_Multi_July2026FundingUpdate/July2026FundingUpdate.md
```diff
@@ -0,0 +1,167 @@
...
+discussions: "https://governance.aave.com/t/direct-to-aip-july-2026-funding-update/25277"
+---
+
+## Simple Summary
+
+This publication presents the July Funding Update, consisting of the following key activities:
+
+- Acquire GHO to support the runway; and
+- Create Allowances to support Operations.
+
```
Kogaroshi, line 16: it seems to be missing some activities here

src/20260715_Multi_July2026FundingUpdate/July2026FundingUpdate.md
```diff
@@ -0,0 +1,167 @@
...
+## Motivation
```
Kogaroshi, line 17: here as well

src/20260715_Multi_July2026FundingUpdate/July2026FundingUpdate.md
```diff
@@ -0,0 +1,167 @@
...
+The MainnetSwapSteward and Aave Finance Committee (AFC) will continue executing a rolling GHO acquisition strategy to maintain adequate runway and preserve sufficient budget to fund ongoing growth initiatives.
+
+### Reimburse Audit Costs
+
+Reimburse TokenLogic for costs incurred in facilitating the audit of one Chainlink PR that was modified to accommodate the GHO ↔ sGHO two-way swap layer. The value of the Audit performed by Trail of Bits was 50,000 aEthLidoGHO.
+
+### Incentive Campaign
+
+To support the continued success of the friendly Tydro deployment, a 5M GHO budget has been discussed with the Tydro team. During June, 2.5M of the 5M GHO was acquired and moved to the Ink Network, ready to support the Kraken DeFi Borrow campaign. The balance will be moved in line with operational requirements.
+
```
Kogaroshi: ?

src/20260715_Multi_July2026FundingUpdate/July2026FundingUpdate.md
```diff
@@ -0,0 +1,167 @@
...
+## Specification
```
Kogaroshi, line 27: missing the StableVault specifications as well

src/20260715_Multi_July2026FundingUpdate/AaveV3Ethereum_July2026FundingUpdate_20260715.sol
```diff
@@ -0,0 +1,194 @@
...
+  address public constant STABLE_VAULT_INCENTIVES_SAFE = 0x4cc755fBFAE135933b64865E7D7Ea6ae778A4D43;
+  uint256 public constant STABLE_VAULT_GHO_ALLOWANCE = 300_000 ether;
+  uint256 public constant STABLE_VAULT_USDC_ALLOWANCE = 100_000e6;
+  uint256 public constant STABLE_VAULT_USDT_ALLOWANCE = 100_000e6;
+
+  address public constant RESCUE_USDC_OWNER = 0x32FcF748e4dCEBD1081BfcccB94eB721101F27C0;
+  uint256 public constant RESCUE_USDC_AMOUNT = 25_000e6;
+
+  uint256 public constant WETH_SWAP_BUDGET_AMOUNT = 5_000 ether;
+  uint256 public constant USDT_SWAP_BUDGET_AMOUNT = 10_000_000e6;
```
avniculae: rlUSD's budget is already exactly 1e24 on-chain and no rlUSD pair has ever been enabled on the steward, so that 1M has never been spendable and this adds a second 1M on top of it. Was the earlier grant meant to be topped up, or is 1M the total the table is asking for? Milder version of the same question for the rest: resulting budgets are WETH 22,460 / USDC 24.9M / USDT 27.87M / USDe 28.9M / USDS 27.36M / DAI 14.8M against 5k/10M/10M/1M/1M/1M in the table.

The May/June write-up spelled out `Method: increaseTokenBudget(token, amount)` and titled the column "Amount"; adding that back would make the intent readable for voters.

src/20260715_Multi_July2026FundingUpdate/AaveV3Ethereum_July2026FundingUpdate_20260715.sol
```diff
@@ -0,0 +1,194 @@
...
+      AaveV3EthereumAssets.RLUSD_UNDERLYING,
+      RLUSD_SWAP_BUDGET_AMOUNT
+    );
+    IMainnetSwapSteward(AaveV3Ethereum.COLLECTOR_SWAP_STEWARD).increaseTokenBudget(
+      AaveV3EthereumAssets.PYUSD_UNDERLYING,
+      PYUSD_SWAP_BUDGET_AMOUNT
+    );
+  }
+
+  function _swapPaths() internal {
```
avniculae, line 121: The table lists GHO, AAVE and wETH as acquirable for every stable, and the forum says unconfigured paths get added here. On-chain `USDe -> AAVE`, `DAI -> AAVE` and `DAI -> WETH` are all false today, and this only adds the rlUSD and pyUSD paths. Should those three be included?

src/interfaces/IMainnetSwapSteward.sol
```diff
@@ -10,4 +14,20 @@ interface IMainnetSwapSteward {
...
+  /// @notice Sets a token pair as allowed for swapping in from -> to direction only
+  /// @param fromToken The address of the token to swap from
+  /// @param toToken The address of the token to swap to
+  /// @param allowed Sets swappable pair to allowed/disallowed
+  function setSwappablePair(address fromToken, address toToken, bool allowed) external;
+
+  /// @notice Sets a token's Chainlink Oracle (in USD)
+  /// @param token The address of the token
+  /// @param oracle The address of the token oracle
+  function setTokenOracle(address token, address oracle) external;
```
avniculae, line 32: This isn't called anywhere. The steward already has oracles set for rlUSD (`0xf0eaC18E...`) and pyUSD (`0x36964C05...`), so `setSwappablePair` is enough - needed for something, or leftover?

src/20260715_Multi_July2026FundingUpdate/AaveV3Ethereum_July2026FundingUpdate_20260715.sol
```diff
@@ -0,0 +1,194 @@
...
+    );
+
+    AaveV3Ethereum.COLLECTOR.approve(
+      IERC20(AaveV3EthereumAssets.WBTC_A_TOKEN),
+      MiscEthereum.AHAB_SAFE,
+      WBTC_MONAD_ALLOWANCE
+    );
+  }
+
+  function _stableVaults() internal {
```
avniculae, line 68: This isn't in the July funding update forum post - that was last edited 16 Jul and the section landed here on 3 Aug. The amounts match section 8 of the Aave App Launch ARFC and that Snapshot passed on 26 Jul (382,932 For / 0.22 Against), so the mandate exists, but the frontmatter says `snapshot: "Direct-to-AIP"` and nothing links the vote. Can we add the Snapshot ID plus a line noting this was added after the forum post?

Separately, `0x4cc755fBFAE135933b64865E7D7Ea6ae778A4D43` appears in neither forum post. On-chain it's a fresh 2/3 Safe (1.4.1, nonce 1, owners `0x9DE1d45e...`, `0x4b752551...`, `0xb291232F...`). Can we get that address confirmed in writing by Labs before this goes on-chain?

src/20260715_Multi_July2026FundingUpdate/AaveV3Plasma_July2026FundingUpdate_20260715.sol
```diff
@@ -0,0 +1,103 @@
...
+contract AaveV3Plasma_July2026FundingUpdate_20260715 is IProposalGenericExecutor {
+  uint256 public constant USDT_ALLOWANCE = 3_000_000e6;
+  address public constant AHAB_SAFE = 0xAA2461f0f0A3dE5fEAF3273eAe16DEF861cf594e;
+
+  function execute() external {
+    _smallAllowances();
+    _refreshUsdtAllowance();
+  }
+
+  function _smallAllowances() internal {
```
avniculae, line 33: Forum says "Amount: Current Balance +1"; this approves exactly `balanceOf` and the md rewrote the line to "Balance at time of execution". These are aTokens, so the balance grows as soon as the tx lands and the allowance is short by the accrual - AFC can't sweep the full position, which is what the section is for. Neither the exact balance nor balance+1 covers that, so it needs a meaningful buffer or a different mechanism.

Two things I hit while checking balances: aPlaPT_USDe_18JUN2026 is 0 on the collector so that approval is a no-op (and its test asserts 0 == 0), and aPlaWETH already has a 2.72 allowance to AFC that this reduces to the 0.26 balance.

src/20260715_Multi_July2026FundingUpdate/AaveV3Arbitrum_July2026FundingUpdate_20260715.sol
```diff
@@ -0,0 +1,87 @@
...
+      MiscArbitrum.AFC_SAFE
+    );
+    AaveV3Arbitrum.COLLECTOR.approve(
+      IERC20(AaveV3ArbitrumAssets.WETH_A_TOKEN),
+      MiscArbitrum.AFC_SAFE,
+      currentAllowanceWeth + WETH_ALLOWANCE
+    );
+
+    AaveV3Arbitrum.COLLECTOR.approve(
+      IERC20(AaveV3ArbitrumAssets.USDCn_A_TOKEN),
```
avniculae, line 58: This one is set absolutely while WETH/USDT/GHO are increments. The allowance is 0 today so it makes no difference now, but if AFC gets an aArbUSDCn allowance before execution this overwrites it while the others add. Deliberate?

src/20260715_Multi_July2026FundingUpdate/AaveV3Arbitrum_July2026FundingUpdate_20260715.t.sol
```diff
@@ -0,0 +1,147 @@
...
+
+  function test_approvals_usdc() public {
+    assertEq(
+      IERC20(AaveV3ArbitrumAssets.USDCn_A_TOKEN).allowance(
+        address(AaveV3Arbitrum.COLLECTOR),
+        MiscArbitrum.AFC_SAFE
+      ),
+      0
+    );
+    _assertAllowanceIncrease(AaveV3ArbitrumAssets.USDCn_A_TOKEN, proposal.USDCn_ALLOWANCE());
```
avniculae, line 87: Adding to your point above: USDCn goes through the same increase helper, but the payload sets that one absolutely, so `allowanceBefore + amount` only holds because the allowance happens to be 0. Better to assert the absolute expected value for this one.

src/20260715_Multi_July2026FundingUpdate/July2026FundingUpdate_20260715.s.sol
```diff
@@ -0,0 +1,223 @@
...
+
+    // register action at payloadsController
+    GovV3Helpers.createPayload(actions);
+  }
+}
+
+/**
+ * @dev Deploy Gnosis
+ * deploy-command: make deploy-ledger contract=src/20260715_Multi_July2026FundingUpdate/July2026FundingUpdate_20260715.s.sol:DeployGnosis chain=gnosis
+ * verify-command: FOUNDRY_PROFILE=deploy npx catapulta-verify -b broadcast/July2026FundingUpdate_20260715.s.sol/8453/run-latest.json
```
avniculae: verify-command points at `/8453/` (Base), should be `/100/`.

src/20260715_Multi_July2026FundingUpdate/AaveV3Plasma_July2026FundingUpdate_20260715.sol
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * @title July 2026 Funding Update
+ * @author TokenLogic
+ * - Snapshot: Direct-to-AIP
+ * - Discussion: https://governance.aave.com/t/direct-to-aip-july-2026-funding-update/25277
+ */
+contract AaveV3Plasma_July2026FundingUpdate_20260715 is IProposalGenericExecutor {
+  uint256 public constant USDT_ALLOWANCE = 3_000_000e6;
+  address public constant AHAB_SAFE = 0xAA2461f0f0A3dE5fEAF3273eAe16DEF861cf594e;
```
avniculae, line 22: Worth adding the explorer link comment like the other hardcoded addresses in this PR have. `MiscPlasma` has no `AHAB_SAFE`, so maybe also worth adding it to the address book since Ethereum/Arbitrum/Gnosis all have it.

avniculae:
On the 25k USDC: I don't think it's recoverable. `AToken.rescueTokens` reverts for the underlying — simulating `rescueTokens(USDC, collector, 25_000e6)` on aEthUSDC from EXECUTOR_LVL_1 reverts with `UnderlyingCannotBeRescued()`, while the same call for any other token goes through. `Pool.rescueTokens` only covers tokens stuck on the Pool itself, and there's no way to sync `virtualUnderlyingBalance` up to the real balance, so it just sits as untracked surplus (aEthUSDC holds 174,408,589.81 USDC against a virtual balance of 174,376,888.13, so ~31.7k of donations in total). The only route I see is an aToken implementation upgrade, which isn't worth it for 25k.

Can we drop `RESCUE_USDC_OWNER` and `RESCUE_USDC_AMOUNT` in that case?

src/20260715_Multi_July2026FundingUpdate/AaveV3Base_July2026FundingUpdate_20260715.sol
efecarranza: sure, will do a follow up PR to add to aave-helpers

src/20260715_Multi_July2026FundingUpdate/AaveV3Ethereum_July2026FundingUpdate_20260715.sol
efecarranza: There were supposed to be some swaps that didn't take place, will remove allowances here and pass onto August/September funding update.

src/interfaces/IMainnetSwapSteward.sol
efecarranza: just adding it as it's usually used so future PRs will use it anyway

src/20260715_Multi_July2026FundingUpdate/AaveV3Plasma_July2026FundingUpdate_20260715.sol
efecarranza: yep, requested to add

efecarranza:
@Kogaroshi @avniculae updated, thanks! 
adam is reviewing the SAFE address for stable vaults incentives, will tag you there

src/20260715_Multi_July2026FundingUpdate/AaveV3Ethereum_July2026FundingUpdate_20260715.sol
efecarranza: https://github.com/aave-dao/aave-address-book/pull/1497

added to address book

src/20260715_Multi_July2026FundingUpdate/AaveV3Plasma_July2026FundingUpdate_20260715.sol
avniculae: The buffer closes the accrual gap, but a flat amount against balances spanning 0 to 7,657 leaves six of the nine allowances larger than the position they cover. From the diff report: aPlaPT_USDe_18JUN2026 gets 100 against a zero balance, aPlaPT_USDe_9APR2026 100.4967 against 0.4968, aPlaweETH 2.1320 against 0.1328.

Once the AFC sweeps, the unused buffer survives as a live allowance and nothing revokes it, so if the collector later receives any of these assets the AFC can pull them with no new proposal. That cuts against a section meant to close these positions out.

Two ways to fix it, and they combine well. A proportional buffer (`balance * 1.001`) tracks accrual at every position size and skips the empty one naturally. And the matching revokes could ride in the next funding update, the way this PR already revokes aArbweETH, aSonwS and the rest from the previous round. The AFC can't revoke for itself, since `COLLECTOR.approve` is funds-admin only, so it has to be a later payload either way.

Separately, the `@dev` line says the split keeps notional headroom comparable across the two classes. At these values it isn't: 100 stable-correlated is roughly $100-124 with sUSDe at $1.2425, against roughly $3,808 for 2 aPlaWETH with WETH at $1,904.05.

One more thing: the forum post was edited on 11 Aug and still reads "Amount: Current Balance +1", so that line probably needs updating there too.

src/20260715_Multi_July2026FundingUpdate/AaveV3Plasma_July2026FundingUpdate_20260715.sol
avniculae: The link points at etherscan for an address only used on Plasma, while the Sonic payload in this PR links sonicscan.org for its own. Also `MiscPlasma.AHAB_SAFE` was merged into the address book on 7 Aug with this exact value, so the constant can go entirely on the next submodule bump.

src/20260715_Multi_July2026FundingUpdate/AaveV3Ethereum_July2026FundingUpdate_20260715.sol
avniculae: The address is now `MiscEthereum.STABLE_VAULT_DAO_WALLET_INCENTIVES` in the address book, so that half is settled and the hardcoded constant can go on the next submodule bump.

The provenance half is still open though. The forum post was edited on 11 Aug, after this section was added here, and it still has no Stable Vaults section, while the frontmatter says `snapshot: "Direct-to-AIP"`. Could we add the Aave App Launch snapshot ID to the write-up so a voter can trace where the $500k mandate comes from?

efecarranza: added snapshot of aave app

src/20260715_Multi_July2026FundingUpdate/AaveV3Plasma_July2026FundingUpdate_20260715.sol
efecarranza: updated here and changed the text of the forum post

efecarranza:
@avniculae updated

Kogaroshi (approved):

DhairyaSethi (approved):

PR #1172, opened 2026-07-23
test: compile generated Solidity in generator tests

AlbertoCentonze:
- add a shared helper that compiles generated payloads, tests, and deployment scripts with Forge
- provision Foundry in the JavaScript CI job
- add tests to make sure that generated solidity code compiles to prevent discovering about failures when the code is compiled

This isolates the remaining generated-bytecode coverage from #1152

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,4 +55,16 @@ jobs:
           exit 1
 
   test-js:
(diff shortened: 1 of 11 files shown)


Kogaroshi (approved):
lgtm for 1st version, we'll have to extend to V4 as well

PR #1175, opened 2026-07-23
feat: Onboard syrupUSDG on Aave V4 Global Dollar Hub

Kogaroshi:
### Pre-review checklist:

- [x] I have run a spell check on the write-up and made sure no typos exist.
- [x] References to Snapshot/governance forum are correct on the AIP. If no snapshot exists, make sure no TODO's exist.
- [x] The specification on the writeup is aligned with the forum, snapshot, and the payload contract. If there are any changes, they are explicitly mentioned/communicated.
- [x] Minimal tests exist, and the snapshot diff report generated is the latest one and aligned with the payload.
- [ ] If deploy scripts are manually updated from the generated ones, I have carefully validated that they are correct, including the deploy commands and the proposal-creation script.
- [ ] If the aave-helpers submodule is updated, I have validated it is pointing to the latest version.
- [x] I have validated that no unused files and imports are being added.
- [x] For an asset listing, the write-up includes a detailed specification of the price feed used, CAPO adapters (with each CAPO layer described separately), and eModes (with tables) if changed.
- [x] Wherever possible, I have validated that addresses from the address book are used instead of raw addresses.

PR #1097, opened 2026-05-14
feat: Launch V4 Paxos Hub and onboard PT-USDG-24SEP2026

avniculae:
## Onboard PT-USDG-24SEP2026 (Paxos Hub)

- Launches a dedicated Paxos Hub with a single USDG Pendle spoke (a correlated stablecoin group)
- Onboards PT-USDG-24SEP2026 as collateral-only on the spoke
- Borrowable set: USDC and USDT supplied natively to the Paxos Hub; USDG drawn from the Core Hub through a cross-hub credit line
- Migrates the USDG price reference — on the Main/Forex/Gold spokes, the new USDG Pendle spoke, and as the PT-USDG-24SEP2026 oracle base — from the fixed $1.00 feed to the capped Chainlink USDG/USD market feed (`PriceCapAdapterStable`, $1.04 cap)

## Notes

- The Security Council will need to register the new spoke on the 4 position managers

https://governance.aave.com/t/direct-to-aip-onboard-pt-usdg-24sep2026-to-aave-v4-on-ethereum/24942

PR #1177, opened 2026-07-24
fix: V4 generator fixes & improvments

### LINKED ISSUE #1174: V4 Generator fixes
- Add AccessManager wiring for Spoke/Hub in AccessManager if 1st listing and not wired yet
- Use percentage inputs instead of bps inputs & convert units into full decimal units & decimals to WAD
- Set chain TreasurySpoke as default feeReceiver
- See for default IR strat ?
- For Hub & Spoke listing => have all the active/paused/frozen/halted to default values
- select multiple PositionManagers to onboard on the same spoke
- when adding a Hub or Spoke via input, ask for a label for them, and use label for rest of prompt (skip if already labeled in previous choices) & for asset inputed, use the symbol + keep memory of previous actions in serie/bundle for labels, Spoke listing on Hubs, etc ...
- add spaces between functions & tests in template & between constants for readability
- add New Spokes, Hubs, assets as constants in the Payload, using their label/symbol as constant
- at the end, before generating, let's check in address-book for that market+chain (& the gov on that chain as well), if any inputed address is known => replace with the address-book entry for it

& add more tests for the payload : 
- inputs & risk params confirmation against expected values (on Hub, Spoke, for asset, caps, IR data)
- tokenization Spoke config, name/symbol & proxyAdmin
- e2e tests for all new Hubs/Spokes/TokenizationSpokes

Kogaroshi:
Fixes #1174


generator/templates/test.template.ts
```diff
@@ -25,7 +25,17 @@ export const testTemplate = (
...
+
+  const helpers = Array.from(
+    new Set(
+      marketConfig.artifacts
+        .map((artifact) => artifact.test?.helpers)
+        .flat()
+        .filter((f): f is string => f !== undefined)
+        .map((h) => h.trim()),
+    ),
+  ).join('\n\n');
```
avniculae, line 62: Only `helpers` gets deduped here — `fn` is a plain concat, so two features contributing the same module each emit their own `test_<feature>Input()` and the generated test file doesn't compile. I emitted the `useCases` fixture to disk and ran forge on it:

```
Error (1686): Function with same name and parameter types defined twice.
  --> ...t.sol:40   function test_hubAssetListingsInput() ...
  Note: Other declaration is here: ...t.sol:130
Error (1686): ... test_hubSpokeToAssetsAdditionsInput() ... :94 / :253
```

`finalizeV4Artifacts` merges `v4Getters` across artifacts so the payload is fine (compiles clean on its own); the test fns never get the same treatment. Feature selection is a checkbox in `cli.ts:180`, so `onboardAssetToHub` + `onboardReserveToSpoke`, or a bundle plus the standalone `hubAssetListing` module, hits it. `test_accessManagerTargetFunctionRoleUpdatesInput` will collide the same way once both bundles have non-empty `targetFunctionRoles`. A single bundle on its own compiles, which is why it isn't obvious.

generator/features/v4/accessWiring.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/// AccessManager wiring granting HUB_CONFIGURATOR_ROLE the HubConfigurator
+/// selectors on a freshly deployed Hub.
+export function hubConfiguratorWiring(hubExpr: string): V4TargetFunctionRoleUpdate {
+  return {
+    target: hubExpr,
+    selectors: [],
+    selectorsExpr: 'Roles.getHubConfiguratorRoleSelectors()',
+    roleId: 'Roles.HUB_CONFIGURATOR_ROLE',
+  };
```
avniculae: `setupHubAllRoles` wires three roles on a fresh hub — HUB_CONFIGURATOR_ROLE, HUB_FEE_MINTER_ROLE and HUB_DEFICIT_ELIMINATOR_ROLE. On-chain all four Ethereum hubs have `addAsset`→101, `mintFeeShares`→102, `eliminateDeficit`→103, but this only emits 101, so a generated fresh hub ships without the fee-minter and deficit-eliminator mappings. Same shape as the spoke 302 gap this PR fixes; 102/103 have no members today so it's latent rather than live.

generator/features/v4/spoke/spokeLiquidationConfigUpdate.ts
```diff
@@ -1,16 +1,31 @@
...
+import {accessorIdentifier, testAddressRef, wrapAddress} from '../testHelpers';
 
-async function sentinelNumber(message: string): Promise<Sentinel> {
-  const v = await numberPrompt({message: `${message} (empty = keep current)`});
-  if (!v) return keepCurrent();
-  return literal(v.replace(/\B(?=(\d{3})+(?!\d))/g, '_'));
+async function sentinelWad(message: string): Promise<Sentinel> {
+  const v = await input({
+    message: `${message} (empty = keep current)`,
+    validate: (x) => x === '' || !isNaN(Number(x)) || 'Enter a decimal number',
```
avniculae: `!isNaN(Number(x))` is looser than the decimal pattern `percentPrompt` uses, and some of what it accepts produces Solidity that doesn't compile. Actual `decimalToWad` output: `1e-3` → `1e-3e18`, `0x10` → `0x10e18`, `Infinity` → `Infinitye18`, `-1` → `-1e18` (compiles, nonsense HF), and `1.0000000000000000009` → `1e18` from the silent `slice(0, 18)`. Reusing `pattern: /^[0-9]*\.?[0-9]*$/` would close all of them. Also this `sentinelWad` is a duplicate of the one in `sentinelPrompts.ts`.

generator/features/v4/bundles/onboardReserveToSpoke.ts
```diff
@@ -38,99 +43,140 @@ type BundleCfg = {
...
+  const t =
+    (await input({
+      message: 'targetHealthFactor (health factor, e.g. 1.05)',
+      validate: (x) => x === '' || !isNaN(Number(x)) || 'Enter a decimal number',
+    })) || '1';
+  const h =
+    (await input({
+      message: 'healthFactorForMaxBonus (health factor)',
+      validate: (x) => x === '' || !isNaN(Number(x)) || 'Enter a decimal number',
+    })) || '1';
```
avniculae: Defaulting an empty health factor to `'1'` gives `targetHealthFactor: 1e18` / `healthFactorForMaxBonus: 1e18` — risk parameters that look deliberate but were never entered. Worth making these required instead?

Separately, `liquidationUpdates` (L244) and `pmUpdates` (L248) aren't guarded by a per-spoke Set the way `wiredSpokes`/`wiredHubs` are, so onboarding two assets to the same spoke pushes duplicate entries for it.

generator/features/v4/hub/hubAssetListingPrompt.ts
```diff
@@ -0,0 +1,70 @@
...
+    message: 'Deploy a TokenizationSpoke for this asset?',
+    default: false,
+  });
+  let tokenization: V4HubAssetListing['tokenization'];
+  if (withTokenization) {
+    tokenization = {
+      addCap: (await numberPrompt({message: 'TokenizationSpoke addCap (whole units)'})) || '0',
+      proxyAdminOwner: await promptProxyAdminOwner(m),
+      name: await input({message: 'TokenizationSpoke name'}),
+      symbol: await input({message: 'TokenizationSpoke symbol'}),
```
avniculae: No validation that name/symbol are non-empty. With the tokenization block entered and `proxyAdminOwner` required, an empty name hits `hasProxyAdminOwner && !hasName` in `HubEngine._deployAndRegisterTokenizationSpoke` and reverts on execution. The generated tests do execute the payload so it'd fail loudly, but a `validate` is cheaper. `esc()` also escapes `'` but not `\\`, so a backslash in either field emits an invalid escape.

generator/features/v4/proxyAdminOwner.ts
```diff
@@ -3,9 +3,12 @@ import * as addressBook from '@aave-dao/aave-address-book';
 import {MarketIdentifierV4} from '../../types';
 import {getMarketChain} from '../../common';
 import {addressPrompt} from '../../prompts/addressPrompt';
+import {resolveEntity} from './entityInput';
 
-/// Prompts for a TokenizationSpoke proxy admin owner. Defaults to the market
-/// chain's Executor LVL1 (with the option to override with a custom address).
+/// Prompts for a TokenizationSpoke proxy admin owner and returns a Solidity codegen
+/// expression. Defaults to the market chain's Executor LVL1 (with the option to
+/// override); a custom address is reverse-resolved against the address book or
```
avniculae, line 10: Should this default to the Protocol Security Council rather than Executor LVL1? Every V4 TokenizationSpoke ProxyAdmin on Ethereum is owned by 0x187AAE17d4931310B3fc75743e7F16Bdc9eD77e9 (checked eight of them), and that's what #1175 settled on. EXECUTOR_LVL_1 is the owner on the USDG Pendle spoke from #1097, which as I understood it was the one we didn't want to repeat — and with `default: true` on the confirm below, pressing enter picks it. Is there a reason the executor is the right default for new deployments?

avniculae:
One thing worth adding while this is open: nothing in CI compiles the generated Solidity. `test-js` runs `vitest generator`, which only diffs strings against `__snapshots__`, and `test-sol` runs with `MODE: "CHANGED"` — this PR touches no `src/**.sol`, so the forge job has nothing to build. That's why the duplicate-test-name break I left inline is green here.

A fixture that writes a generated payload+test into a temp `src/` dir and runs `forge build` on it would cover it. That's how I hit it — emitting the `useCases` config to disk and building.

generator/features/v4/proxyAdminOwner.ts
Kogaroshi: I decided to have it as default as this is the aim for the final ownership configuration after the hardening phase. The current behavior to set Protocol Security Council is more of a temporary override, which should be inputed when prompted for now.

generator/features/v4/hub/hubAssetListingPrompt.ts
Kogaroshi: 05d7b9f5a5c13483becf607abe07c7ce0fcb7867

generator/features/v4/bundles/onboardReserveToSpoke.ts
Kogaroshi: 05d7b9f5a5c13483becf607abe07c7ce0fcb7867

generator/features/v4/spoke/spokeLiquidationConfigUpdate.ts
Kogaroshi: 05d7b9f5a5c13483becf607abe07c7ce0fcb7867

generator/features/v4/accessWiring.ts
Kogaroshi: 05d7b9f5a5c13483becf607abe07c7ce0fcb7867

generator/templates/test.template.ts
Kogaroshi: 05d7b9f5a5c13483becf607abe07c7ce0fcb7867

generator/features/v4/hub/hubSpokeToAssetsAddition.ts
```diff
@@ -2,9 +2,31 @@ import {checkbox, confirm} from '@inquirer/prompts';
...
+/// an N-asset array instead of N items.
+function aggregateByHubSpoke(cfg: V4HubSpokeToAssetsAddition[]): V4HubSpokeToAssetsAddition[] {
+  const merged = new Map<string, V4HubSpokeToAssetsAddition>();
+  for (const c of cfg) {
+    const key = `${c.hubLib}|${c.spoke}`;
+    const existing = merged.get(key);
+    if (existing) existing.assets = [...existing.assets, ...c.assets];
+    else merged.set(key, {...c, assets: [...c.assets]});
+  }
+  return [...merged.values()];
```
avniculae: nit: this merges within a single `build` call, so two features both registering assets on the same (hub, spoke) still emit two entries — the useCases fixture output has two `SpokeToAssetsAddition` for MAIN_SPOKE, one asset each. Harmless, the payload is correct either way and it's the same shape as the two `addSpoke` calls #1175 makes for assetIds 3 and 4, just not as tidy as the single-feature case.

Kogaroshi: 581641fbba2bc5d1723803afbb863549e34fca75

avniculae (approved):

PR #1178, opened 2026-08-03
fix : apply aave-helper changes & refactor generator template tests

Kogaroshi:


generator/features/v4/accessWiring.ts
```diff
@@ -1,95 +1,52 @@
...
-      selectorAsserts: HUB_CONFIGURATOR_SELECTORS,
-    },
-    {
-      target: hubExpr,
-      selectors: [],
-      selectorsExpr: 'Roles.getHubFeeMinterRoleSelectors()',
-      roleId: 'Roles.HUB_FEE_MINTER_ROLE',
-      selectorAsserts: ['IHub.mintFeeShares.selector'],
+  const hubTests = entities.hubs.map((hub) =>
+    rolesWiredTest(accessorIdentifier(hub), testAddressRef(hub), 'address(0)'),
```
avniculae: Should hubs get a reference target too? Spokes get `referenceSpoke(market)` on the line below, but hubs pass `address(0)`, so a fresh hub gets no divergence check against an already-wired one. `CORE_HUB` would work — I checked, all four Ethereum hubs carry 101/102/103 for those selectors — and it's one line.

Defensible to leave as-is, since the equivalence test against `setupHubAllRoles` in helpers is a stronger guard than a reference comparison. Just that the hub roles were the gap that started this, so the asymmetry stood out.

generator/features/v4/bundleHelpers.ts
```diff
@@ -20,37 +20,65 @@ export function finalizeV4EntityConstants(
...
+/// Fold array expressions into nested `arrayMerge` calls, as the helpers library
+/// concatenates two arrays at a time.
+const foldArrays = (parts: string[], arrayMerge: string) =>
+  parts.reduce((acc, part) => `${arrayMerge}(${acc}, ${part})`);
+
+/// Body of a merged getter: the entries it collected, then any library-built arrays
+/// appended after them, so the entry indices the input assertions use hold.
+function getterBody(value: V4GetterEntry): string {
+  const arrays = value.arrayExprs ?? [];
+  if (value.entries.length === 0) return `return ${foldArrays(arrays, value.arrayMerge!)};`;
```
avniculae, line 38: nit: `foldArrays` reduces without an initial value and this asserts `arrayMerge!`, so a getter arriving with `arrayExprs` but no `arrayMerge`, or with both `entries` and `arrayExprs` empty, fails as a raw TypeError rather than something readable. Unreachable today — `accessWiringArtifact` is the only producer and sets both, and both bundles guard on `freshHubs.length > 0 || freshSpokes.length > 0` — so just a note for the next caller.

Kogaroshi: 8515950d0585b397b69f6f52a31f5155cdbeb6eb

generator/features/v4/accessWiring.ts
Kogaroshi: 8515950d0585b397b69f6f52a31f5155cdbeb6eb

generator/features/v4/fixtures.ts
```diff
@@ -354,7 +362,8 @@ export function useCasesFixture(): Fixture {
   };
 
   const onboardAssetCfg = {
-    targetFunctionRoles: hubWiring(HUB),
+    freshHubs: [HUB],
```
avniculae: Using book entities as the fresh ones makes the new divergence check vacuous: `referenceHub` resolves to CORE_HUB, which is also the target, so the generated test reads `_assertRolesWired(items[i], address(AaveV4EthereumHubs.CORE_HUB))` and compares the target's roles against its own. Same on line 405 with MAIN_SPOKE.

Real payloads can't hit it — a fresh hub or spoke is by definition not in the book — but the fixture is the only place this codegen path gets compiled, so a wrong `referenceHub`/`referenceSpoke` wouldn't show up. Worth pointing `freshHubs`/`freshSpokes` at a raw address instead?

Kogaroshi: 29dfe772ceb3b8e4a3c933e4e026705b925d359d

avniculae (approved):

PR #1182, opened 2026-08-17
fix: default addCap for non-borrowable & bump libs

Kogaroshi:
Set default TokenizationSpoke addCap for non-borrowable listing to 0 during payload generation & bump dependencies


DhairyaSethi (approved):

foundry.lock
```diff
@@ -2,13 +2,10 @@
   "lib/aave-helpers": {
     "branch": {
       "name": "main",
-      "rev": "1165ff370b1efcd6c8c3f5340e4680cef1b96400"
+      "rev": "4e828a86159dbf604deb719e44c4c09c36784ec6"
     }
   },
   "lib/aave-umbrella": {
-    "tag": {
-      "name": "v1.0.0",
```
DhairyaSethi, line 10: why were they on this tag before?

DhairyaSethi: checked the commit, its fine

Kogaroshi: from bgd repo iirc

PR #1188, opened 2026-08-20
fix: July 2026 Funding Update proposal md corrections

DhairyaSethi:
Post-merge corrections to the July 2026 Funding Update proposal text (#1164), before proposal creation pins it to IPFS:

- **Remove the snapshot link** from the frontmatter and replace the reference hyperlink with `Direct-to-AIP`. The linked snapshot (`0x2f860…9280b`) belongs to the [ARFC] Aave App Launch vote, not this proposal — presenting it as this proposal's own snapshot is misleading since this is a Direct-to-AIP submission.
- **Correct the Plasma allowance buffer wording.** The md described a fixed-unit buffer (100 units for stable-correlated assets, 2 for aPlaWETH/aPlaweETH), but the payload applies a 0.10% buffer on the balance (`percentMul(100_10)`). The text now matches on-chain behavior.
- **Restore `blob/main/` permalinks** in the References section so the ipfs.yml workflow re-pins them to the new commit sha on merge (they currently point at `dc5eb0b8`, which predates these text fixes).


DhairyaSethi:
Superseded by #1189.

